PR #75, opened 2021-01-13
circle annotation

AlaEmrani:
I created a circle annotation tool. it added in tools/circle directory and works with CircleFragment which defined in selectors directory. It would be called by DrawingTool id 'circle'.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,11677 @@
 {
   "name": "@recogito/annotorious",
-  "version": "2.2.0",
-  "lockfileVersion": 1,
+  "version": "2.2.1",
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@recogito/annotorious",
+      "version": "2.2.1",
+      "license": "BSD-3-Clause",
+      "dependencies": {
+        "@babel/polyfill": "^7.10.4",
+        "@recogito/recogito-client-core": "^0.3.2",
+        "axios": "^0.19.2",
+        "preact": "^10.4.1",
+        "tiny-emitter": "^2.1.0"
+      },
+      "devDependencies": {
+        "@babel/core": "^7.6.2",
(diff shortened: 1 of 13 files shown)


rsimon:
That's excellent - thanks! The PR could use a little cleanup before merging though. Can you perhaps submit a request that contains only the new tool files? And not, e.g., the formatting changes in unrelated files?

Two more questions: 

- I wasn't aware the media fragments standard has a "circle" syntax. Is this standard? I think it would be necessary to serialize the circle as an SVG shape instead, in order to make it compliant with the W3C Web Annotation specification.
- The mask feature is currently unimplemented for the circle, right?

AlaEmrani:
You're welcome. In addition to adding the circle tool and a circleFragment in the selector directory, I changed the setcurrent in DrawingTolls to circle (which I will revert it in new merge request), added drawing function in the selector/index.js and change 'a9s-inner' and 'a9s-outer' fill attribute in CSS file which not affected other tools.

- no, it's not a standard syntax, I just made it up, I would gladly refine it if you could show me the way.
- yes, I couldn't figure out the usage of the mask element and its 'd' attribute format, if you could explain it, I will correct it too.

rsimon:
Sounds great - thanks again for this. Ok, I'll wait for that revert, and will then merge.

I need to see how to best serialize the SVG circle shape. It should probably done in a similar way as for the polygons right now. But don't worry about it, I can handle that later myself, too.

Concerning the mask: it's possible somehow using SVG path syntax, by creating a circular arc path, or possibly two semi-circular ones. Probably requires a bit of experimentation. But no worries about that either. Again, I can also look into that later. Meanwhile, not having a mask is just a cosmetic issue. Nothing that breaks the tool & it's fine the way it is.

AlaEmrani:
Thank you for the review and comments. I would work on a point annotation tool in following. I hope it would be useful.

rsimon:
Just tested it and it's great - thx! Yes, a point annotation tool would be very useful, too.

PR #79, opened 2021-02-10
support w3c annotations using <path> instead of <polygon>

ap--:
Hello everyone,

I am currently playing around with the annotorious openseadragon plugin, and I noticed that chrome couldn't set the fill opacity for the loaded annotations. I then realized that my w3c annotations use `<path>` instead of `<polygon>` and so the base css for annotations in annotorious wasn't applied. 

This PR fixes the issue where w3c annotations using the SvgSelector with `<path>` elements instead of `<polygon>` are not drawn with transparent fill-color.

Cheers,
Andreas 😃

### DIFF
--- a/src/ImageAnnotator.scss
+++ b/src/ImageAnnotator.scss
@@ -25,7 +25,7 @@
 
 .a9s-annotation, .a9s-selection {
 
-  rect, polygon {
+  rect, path, polygon {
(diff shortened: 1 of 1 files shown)


rsimon:
Ah - yes, that's very useful. Many thanks!

PR #130, opened 2021-06-18
Adds support for circle selection

Hero-Solutions:
A pull request was made on 14 January 2021 to add support for circle selection. However, the codebase has undergone a lot of changes since and the circle selection feature is no longer compatible with the current revision of the main branch. This pull request fixes that.

This may seem like a pretty big commit, but the changes compared to the main branch are relatively small. To check the actual difference between the main branch and the new feature-circle-selection branch, you can check out https://github.com/Hero-Solutions/annotorious/compare/4c439cec30bcd62aa4f4e79b87b801c0a4711582...89119fb36d2e0ec5db9cb67ff0135e7e3a121b2e


rsimon:
Many thanks for this!! I'm margin into the main branch now. However just a note: I'll refactor this, because I don't want to add additional drawing tools to the core distribution anymore, to keep the base bundle size small.

Instead, I'm going to start a separate plugin (`annotorious-selector-pack`) which will collect additional useful selection tools. I'll move the circle tool there as soon as the project is set up.

Two additional thoughts, maybe you have ideas or an opinion about his:

- the way circles are stored isn't standard compliant right now, I think. AFAIK there is no CircleFragment selector in the Media Fragments specifications. This means the circle would have to be stored in the annotation with an SvgSelector, in a similar way as the [Polygon](https://github.com/recogito/annotorious/blob/main/src/selectors/EmbeddedSVG.js#L90-L105)
- The dim mask (the optional SVG element which darkens the image outside the selection while editing) isn't currently implemented if I saw correctly? I wonder how to do this for the circle shape. We'd probably need a circle-shaped SVG `path` element for this I guess.

Hero-Solutions:
You're most welcome. I'm currently working out two more selectors (ellipse and free drawing), but I won't be making pull requests for them just yet.

Let me know when the new selector pack plugin is set up, then I'll see if I can properly integrate the ellipse and free draw shapes into it, as long as it doesn't prove too challenging.

Thank you for reviewing the PR. I wasn't aware that a CircleFragment isn't standard compliant. I had also overlooked implementation of the dim mask. I'll see what I can do.

PR #132, opened 2021-06-19
Feature circle selection

rsimon:


rsimon:
I'm now merging this into the main branch - thanks so @AlaEmrani and @Hero-Solutions for these additions, I really appreciate that!

I'm just reposting a bit of extra info from @Hero-Solutions last PR:

---

 I'll refactor this, because I don't want to add additional drawing tools to the core distribution anymore, to keep the base bundle size small.

Instead, I'm going to start a separate plugin (`annotorious-selector-pack`) which will collect additional useful selection tools. I'll move the circle tool there as soon as the project is set up.

Two additional thoughts, maybe you have ideas or an opinion about his:

- the way circles are stored isn't standard compliant right now, I think. AFAIK there is no CircleFragment selector in the Media Fragments specifications. This means the circle would have to be stored in the annotation with an SvgSelector, in a similar way as the [Polygon](https://github.com/recogito/annotorious/blob/main/src/selectors/EmbeddedSVG.js#L90-L105)
- The dim mask (the optional SVG element which darkens the image outside the selection while editing) isn't currently implemented if I saw correctly? I wonder how to do this for the circle shape. We'd probably need a circle-shaped SVG `path` element for this I guess.

PR #137, opened 2021-06-21
Adds support for freehand/path drawing tools

Hero-Solutions:
Unlike a polygon, there is no "points" array attached to the shape element to calculate the area, so we have to rely on splitting and parsing the "d" attribute instead.

The main concern with this approach is that a path can consist of a large amount of points, potentially impacting performance. Any input to improve performance is more than welcome.


rsimon:
Thanks for this! 

I wouldn't worry too much about the performance for now. The same problem applies to polygons, too. Although you are right, of course, that users are likely to produce (way...) more points when drawing freehand, than when creating polygons. One solution (further down the line) might be to apply some kind of polygon simplification to the path, after the user is finished drawing. The area computation as such probably can't be optimized much I assume.

PR #142, opened 2021-07-05
Adds a setWidgets method to the Annotorious class.

lauxley:
Allows to do something like that:
```js
   this.anno.disableEditor = true;
   this.anno.setWidgets([]);
   this.anno.setWidgets(['COMMENT']);
```
My editor removed a lot of pending spaces, unfortunately it creates a lot of noise even though there are very few changes.


rsimon:
Great - many thanks! Yes, the conflict was about a bug I fixed yesterday night & quick to resolve.

PR #149, opened 2021-07-20
Prevents right clicks from triggering all behaviors.

lauxley:
I tried to prevent right clicks everywhere but there is at least one behavior I couldn't find: adding points to a polygon (guessing it's somewhere in EditablePolygon.js but can't find it).  

I extended `EditableShape` from Tool to use `getSVGPoint` and it felt right but maybe it may have side effects?

### DIFF
--- a/src/AnnotationLayer.js
+++ b/src/AnnotationLayer.js
@@ -87,7 +87,7 @@ export default class AnnotationLayer extends EventEmitter {
 
     elem.addEventListener('mouseleave', () => {
       if (!this.tools?.current.isDrawing) {
-        if (this.currentHover === elem) 
+        if (this.currentHover === elem)
           this.emit('mouseLeaveAnnotation', annotation, elem);
 
         this.currentHover = null;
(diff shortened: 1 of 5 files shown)


rsimon:
Thanks for this! I think you might be right about making EditableShape extend from Tool... I'll need to double check if there are any unwanted side effects. If not, this definitely eliminates some redundancy. 

I also checked quickly about adding polygon points. It all first processed in this handler function in the Tool base class:
https://github.com/recogito/annotorious/blob/main/src/tools/Tool.js#L74

My guess is if you add the button check there, it might not be necessary at various other places... But as I said, I only took a very quick look now.

lauxley:
Ok I'll try to update `Tool.MouseUp` see if it allows to clean up some of the checks.

lauxley:
I had to leave all the checks, if I remove the one from `AnnotationLayer._onMouseDown` then drawing is possible, and removing for either `onGrab` make it possible to drag with right clicks.  
I don't have a smartphone to test the solution, does the check works on all devices?

rsimon:
Thanks a ton for this!! Yes, I re-read the code yesterday night and it looks like the checks are necessary in all those places. I.e. your PR was almost perfect the first time already, just missing the check in that one method. I'll try to test this out tonight and will merge then.

Cheers,
R

PR #186, opened 2021-11-26
Fix getSelectImageSnippet usage

bricesanchez:
before this function returned "undefined", now we have the cropped snippet from the bounding box

before the fix:
annotorious-before-fix-getselectedimagesnippet

after the fix:
annotorious-after-fix-getselectedimagesnippet

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -83,8 +83,10 @@
           console.log('changed from', previous, 'to', selected);
         });
 
-        anno.on('createAnnotation', function(a) {
-          console.log('created', a);
+        anno.on('createAnnotation', function (annotation) {
+          const { snippet, transform } = anno.getImageSnippetById(annotation.id);
+          console.log(snippet.toDataURL('image/jpeg'));
+          console.log('created', annotation);
         });
 
         anno.on('updateAnnotation', function(annotation, previous) {
--- a/src/ImageAnnotator.jsx
+++ b/src/ImageAnnotator.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import AnnotationLayer from './AnnotationLayer';
 import { Editor } from '@recogito/recogito-client-core';
+import { getSnippet } from './util/ImageSnippet';
 
 import './ImageAnnotator.scss';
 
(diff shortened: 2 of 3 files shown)


rsimon:
Thanks for this, I'll try to review over the weekend. Just one comment in advance though: this introduces a breaking change to the API. 

A method to grab the snippet for any specific annotation would be useful. But the current API method does not take an argument - it returns the snippet for the current selection.

rsimon:
Hi again,

I got a chance to review this now. Ok, there are two things:

1. There was one bug in ImageSnippet.js which you caught - thanks for spotting & fixing!
2. However, there's confusion about how the `getSelectedImageSnippet()` method is supposed to work. (No doubt also because of my docs not being very detailed...)

About `getSelectedImageSnippet`: the method essentially serves as a companion to `getSelected`, which is to get the data from the current selection. In other words: this is supposed to work (and now works - after your fix to ImageSnippet.js):

```js
anno.on('selectAnnotation', function(a, shape) {
  console.log(anno.getSelectedImageSnippet());
});

anno.on('createSelection', function(a, shape) {
  console.log(anno.getSelectedImageSnippet());
});
```

__However:__ what you tried to do does not work, by design:

```js
anno.on('createAnnotation', function(a, shape) {
  // Returns null, because after the createAnnotation event, there is no more selection!
  console.log(anno.getSelectedImageSnippet());
});
```

That being said: I think it would be super useful to have a new API method that does what you implemented, which is to get the image snippet for a specific annotation. This would then be, essentially, a companion method to [getAnnotationById](https://github.com/recogito/annotorious/blob/main/src/index.jsx#L165-L168) (and should maybe be called `getImageSnippetById`?)

Would you be up to provide a modified PR along these lines?

Cheers & thanks again,
Rainer

bricesanchez:
Awesome @rsimon ! Thanks for your feedback, I've updated my PR based on your comments!

rsimon:
Merging this now - many thanks!!

bricesanchez:
You're welcome! Thank you for your help!

PR #211, opened 2022-03-08
Snippet use outer rect instead of whole annotation element

dbil25:
the methods `.getImageSnippetById` and `.getSelectedImageSnippet` did not return the same result when an annotation is in edit mode (because the element includes the handles).

This was very annoying for cases when we want to show the user a preview of what he is selecting.

for example, when editing this:
image

it returns this image:
image

but once the user press OK it returns this:
image

### DIFF
--- a/src/util/ImageSnippet.js
+++ b/src/util/ImageSnippet.js
@@ -1,14 +1,13 @@
 import { hasClass } from './SVG';
 
 export const getSnippet = (image, element) => {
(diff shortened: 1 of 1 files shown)


rsimon:
Thanks again for this. Well spotted & very nice pull request!

Now looking at it, I think this could actually be simplified even more. (You could always do `querySelector('a9s-inner')`. There's no need for checking if it's an editable or normal annotation. I might test that over the weekend...)

PR #225, opened 2022-05-05
fix error on deletion of new created selection @recogito/annotorious#221

AlexMili:


rsimon:
Hi, can you provide a bit of explanation around this? I'm not sure this really changes anything. If anything, it might be less robust, because e.g. a value of `this.selectedShape == undefined`  would pass through (leading to a null exception in the next line).

AlexMili:
I didn't catch yet the inner working of the lib but this is where an error was raised when I tried to create a new selection after deleting the previous one. I just ran some tests and I think this is related to the `allowEmpty` option combined with `disableEditor` option. Sorry I don't have more clue for the moment. Anyway, checking with `null` instead of `undefined` is not a problem as long as it is compared with `==` and not `===`.

AlexMili:
I ran some more tests with the code below and your current main branch and here is what I got:
```javascript
var anno = Annotorious.init({
  image: 'hallstatt',
  locale: 'auto',
  allowEmpty: false,
  disableEditor: false,
  drawOnSingleClick: true,
});

anno.setDrawingTool('polygon');

document.addEventListener('keyup', (event) => {
  if (event.key == "Backspace" || event.key == "Delete") {
      var selected = anno.getSelected();
      if (selected != null) {
          anno.removeAnnotation(selected);
      }
  }
});
```

- `disableEditor: false, allowEmpty: false`
Everything works as expected.
- `disableEditor: true, allowEmpty: false`
The selection works but once I click outside of it after I finished it, it disappears. I need to add this code (as in your example):
```javascript
anno.on('createSelection', async function(selection) {
  selection.body = [{
    type: 'TextualBody',
    purpose: 'tagging',
    value: 'MyOtherTag'
  }];

  await anno.updateSelected(selection);

  anno.saveSelected();
});
```
Now with this code, I get `t.selectedShape is null` error when I double click to finish the selection but the annotation is created and is persisted. I can then delete it without problem, but of course the same error is raised again when I try to create another selection/annotation but again it is persisted anyway.

- `disableEditor: true, allowEmpty: true`
Selection/Annotation works fine, I can click outside of it after finishing it, no error is raised. I can also delete it without problem. The `t.selectedShape is null` is raised when I try to create a new annotation after deleting the first one but it is not persisted, it is only persisted when I try to add a third annotation. All selection created after the first one are persisted when I create n+1 selection. Now when I add the previous listener on `createSelection` event, the previous problem on deletion disappears but I have the same error as previously which is an error when I double click to finish the selection.

With my fix on top of yours all the above situations run smoothly. It works even without the `createSelection` handler when `disableEditor: true`.

Maybe there is a deeper problem with how selections and annotations are handled, but I hope my explanations are helping.

trioangle27:
Hi i am using annotorious plugin.
 mobile is not working plugin. is possiple mobile responsive this plugin ?

PR #275, opened 2023-02-08
Add config option to enable resizing of rects via edges

derek-young:
This PR adds a property to the config object, `enableEdgeControls`, that defaults to `false` but when enabled allows for adjusting an Annotorious EditableRect via its edges. This is in response to a recent discussion: https://github.com/recogito/annotorious/discussions/274

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -60,6 +60,7 @@
           image: 'hallstatt',
           locale: 'auto',
           drawOnSingleClick: true,
+          enableEdgeControls: true,
           widgets: [
             { widget: 'COMMENT' },
             { widget: 'TAG', vocabulary: [ 'Animal', 'Building', 'Waterbody'] }
--- a/src/tools/rectangle/EditableRect.js
+++ b/src/tools/rectangle/EditableRect.js
@@ -11,6 +11,9 @@ import {
   setRectMaskSize
 } from '../../selectors/RectFragment';
 
+const CORNER = 'corner';
+const EDGE = 'edge';
+
 /**
  * An editable rectangle shape.
  */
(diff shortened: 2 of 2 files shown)


vidya-talapady-percipient-ai:
[simon](https://github.com/rsimon)

rsimon:
Thanks for this PR. I still need to find some time to test it. But at first glance, this looks great!

Only one thing: I think this is generally useful to everyone. Therefore I don't think it would even have to be a configurable option. I'd just make it the default. (I would probably set the next version number to 2.8.0 rather than 2.7.11, due to the behavior change.) I'll probably make that change after testing & merging - but feel free to amend the PR, if you want and have the time.

Cheers,
Rainer

derek-young:
Sounds good @rsimon. The reason I thought perhaps it should be configurable is because the handles can be somewhat large relative to the size of the label, so depending on how small the user's labels are they may not want the additional clutter of the edge handles? Not sure... 

Let me know what you think after you've gotten a chance to test, if you'd still like this to be the default behavior I'm happy to make that change.

rsimon:
Thanks - merging this now. I agree though that it makes sense to keep it a configurable option. I only realized just now when testing, that you made edges draggable via extra circle handles. I think I would have done it in a different style, making the edges themselves draggable. (Possibly by adding transparent SVG lines above the edges, a few pixels wide, so that they are easier to grab.) You could also use CSS to change the cursor to the appropriate cursor style (`ew-resize` or `ns-resize`). But this version is definitely a helpful start!

darshanvss:
how do I use the edge handles? I have tried making the same changes in my node module files following the commits as it's still showing 2.7.10 as latest, but I am not able to get the annotations have the edge handles. Should I use the EditableRect using the addDrawingTool and setDrawingTool? I am trying to implement it in VueJS.

vidya-talapady-percipient-ai:
> Thanks - merging this now. I agree though that it makes sense to keep it a configurable option. I only realized just now when testing, that you made edges draggable via extra circle handles. I think I would have done it in a different style, making the edges themselves draggable. (Possibly by adding transparent SVG lines above the edges, a few pixels wide, so that they are easier to grab.) You could also use CSS to change the cursor to the appropriate cursor style (`ew-resize` or `ns-resize`). But this version is definitely a helpful start!

@rsimon here is my branch for the above changes you described. https://github.com/vidya-talapady-percipient-ai/annotorious-2.7.13-custom/commit/de23374faf9242952a95ff11f517f9f0bdbe72c6

PR #279, opened 2023-02-24
Config option to add a className to the container div

zzhong31:
Found this useful when I wanted to change the overall container to flex and adjust the z-index


zzhong31:
@rsimon Any thoughts on this one? There was a particular case where I needed to adjust the css of the overall container and didn't want to modify the DOM directly from outside of the package

PR #291, opened 2023-08-12
bound creation of rectangle to the image

andrheapaltao:
- added constrain in creating rectangle to prevent creation of annotation outside the image


rsimon:
Thanks for this!

PR #292, opened 2023-08-20
Update GitHub links in README

AndrewADev:
Changes to update GitHub links to point directly to new project location.

These were still pointing to a location in the recogito organization, which lead to a nearly blank page before being redirected to the current location (GH Pages links).

As far as I can tell, the other links in the README seem to be correct (e.g. for Gitter)

Note that there is another link to the previous GitHub page in the repository details section (upper right hand corner). Someone with the appropriate permissions will have to update that: 
image


rsimon:
Thanks for catching & fixing those! I'll update the Github link to previous page right away.

PR #294, opened 2023-08-27
GitHub URLs in package.json point to current project

AndrewADev:
Update a few URLs in the `package.json` that were pointing to the project's previous location.

At the time of #292, I didn't think to search for other GH references to recogito throughout the codebase - sorry about that! This time I did a search on GitHub, and there don't look to be any other references relating to GH links (only imports, [naming conventions](https://github.com/annotorious/annotorious/blob/main/webpack.polyfills.js#L4)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/recogito/annotorious.git"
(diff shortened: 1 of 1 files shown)


rsimon:
Good catch - thanks!

PR #306, opened 2023-10-12
Account for viewer offset in setPosition

abiswas97:
`OpenSeadragonPopup`'s setPosition function call would always be relative to window instead of the OSD container. Added an offset to position the popup based on the viewer's position.


rsimon:
Thanks for this! 🙏 Actually now that I look at my code it: it's pretty obvious that I've done all my test cases with full-screen OSD viewers only 😃. 

I know you're using this with v2 currently (and probably as source code directly). But 'm publishing a new v3 release to the npm registry today, in case it helps.

Cheers & thx,
Rainer

Issue #322, opened 2023-12-13
Fix possible `undefined` reference exception on `selector` in `parseW3CImageAnnotation`


oleksandr-danylchenko:
## Issue Description
When the `W3CAnnotation` gets parsed into the `Annotation`, an `error` can be returned from the `parseW3CImageAnnotation`. It happens when no required selectors were found in the spec annotation definition, when `selector` is `undefined`:
https://github.com/annotorious/annotorious/blob/c525576ae44c978deeb902221a7042ac899190f9/packages/annotorious/src/model/w3c/W3CImageFormatAdapter.ts#L45-L59
So when the `Error(Unknown selector type: ${selector.type})` expression executes, the `type` prop is being accessed on the `undefined` `selector`

## Possible Solutions
The `selector` in the problematic expression should be replaced with `w3cSelector`

## Assignees
I would like to start working on this change myself and then submit the PR for review

PR #323, opened 2023-12-13
Fixed possible `undefined` reference exception on `selector` in `parseW3CImageAnnotation`

oleksandr-danylchenko:
## Issue
This PR resolves the https://github.com/annotorious/annotorious/issues/322 issue

## Changes Made
In this PR I replaced accessing the `undefined` `selector` object on error creation with the `w3cSelector`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
+.idea
 dist
 node_modules
 **/config.json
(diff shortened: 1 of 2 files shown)


rsimon:
Yikes, it's never nice when an error message causes an error :-D Thanks for spotting!

Issue #325, opened 2023-12-14
Why the annotation lifecycle properties of the `Annotation` are present on the `AnnotationTarget`?


oleksandr-danylchenko:
Could you clarify why are the `created`, `creator`, `updatedBy`, and `updated` specified on the `AnnotationTarget` and not the `Annotation` itself?...
According to the spec, only the parent annotation can have those, but not the target 🤔 
https://www.w3.org/TR/annotation-model/#lifecycle-information

https://github.com/annotorious/annotorious/blob/4a36371ea227973d72aaf93ca79980a314b30597/packages/annotorious-core/src/model/Annotation.ts#L19-L33

rsimon:
I don't think the spec explicitly prohibits them. Either way: that's the internal model after all. The reasoning behind it is that targets and bodies can be created pretty much independently (think of multiple shapes, and comments and replies). The annotation itself is really just a container. 

In that sense, I've always thought of the `created` / `creator` info at the annotation level as a piece of derived information - whoever was the first creator of the first part of the annotation. In addition, realtime systems will need to push around targets and bodies independently from the rest of the annotations (hence the `annotation` field to connect them). And IMO it's a lot better to keep that info together rather than spread out. (For as similar reason, Annotorious also doesn't use `motivation` at the annotation level. In general, annotations can be lists of bodies all with different `purposes` (tagging, commenting, replying etc.)

oleksandr-danylchenko:
> The reasoning behind it is that targets and bodies can be created pretty much independently

Aha, I'm on the same page now. But it might mean that the existing W3C adapter should be reworked a bit to prevent losing those properties. Because now on parsing those lifecycle props are not being assigned to the `target` and on serialization, they are not lifted from the `target` to the `Annotation` level

rsimon:
Yes, I agree. That bit should probably be revised, for better spec compatibility.

oleksandr-danylchenko:
PR is on the way 🚀

rsimon:
I think this one's now good to close, with your latest PRs merged, right?

oleksandr-danylchenko:
Yeah, now we have the lifecycle props preserved. Thanks 🙌🏻

rsimon:
Thanks for your contribution! BTW: I haven't forgotten on the PRs on Text Annotator. But, frankly, Text Annotator is a bit lower on my priority list, as far as my current running projects are concerned. So it may take a while until I get to that. But thanks in advance already ;-)

PR #326, opened 2023-12-14
Fixed losing the `lifecycle information` on `parsing`/`serialization` of the `W3CAnnotation`

oleksandr-danylchenko:
## Issue
This PR resolves the https://github.com/annotorious/annotorious/issues/325#issuecomment-1855701381 issue

## Changes Made

In this PR I fixed losing the [lifecycle information](https://www.w3.org/TR/annotation-model/#lifecycle-information): 
- when the W3C annotation gets parsed. That was happening because by spec the info lies within the root of the annotation object, but in the inner model it's convenient to co-locate it with the `AnnotationTarger` ([see](https://github.com/annotorious/annotorious/issues/325#issuecomment-1855697146)). So that info wasn't transferred down from the w3c root to the new target.
- when the inner model gets serialized. No method could bring up the lifecycle info from the target to the root of the w3c anno

## Related PRs or Issues
This PR is required for implementing the default W3C Format Adapter for the Text Annotator - https://github.com/recogito/text-annotator-js/pull/7


oleksandr-danylchenko:
Okay, not entirely ready for review...
The `W3CSelector` type property is not compatible with the `W3CTextAnnotationSelector` type that won't have the `value` 🤔 
image
image

packages/annotorious-core/src/model/W3CAnnotation.ts
```diff
@@ -42,11 +58,11 @@ export interface W3CAnnotationBody {
...
 
-export interface W3CAnnotationTarget {
+export interface W3CAnnotationTarget<S = W3CSelector> {
 
   source: string;
 
-  selector?: W3CSelector | W3CSelector[];
+  selector?: S | S[];
 
 }
```
oleksandr-danylchenko, line 67: Made `selector` type generic to support selectors that don't have a `value`, like the `Text Quote Selector` or `Text Position Selector`. Otherwise, we'll encounter the next error - https://github.com/annotorious/annotorious/pull/326#issuecomment-1855987187

oleksandr-danylchenko:
Oh-oh... There's another missed thing. The type of lifecycle info props is `Date`. This is not the case for the W3C format - they are `string`s. So there's a missing conversion step between the `string -> Date` and vice versa

packages/annotorious-core/src/model/W3CAnnotation.ts
```diff
@@ -97,22 +113,58 @@ export const parseW3CBodies = (
...
+  } as unknown as A);
+
+export const serializeW3CTarget = <A extends AnnotationTarget, T extends W3CAnnotationTarget<T['selector']> = W3CAnnotationTarget>(target: A, source: string, selector: T['selector']): T => {
+  const { creator, created, updatedBy, updated, ...serializableRest } = target;
+  return {
+    ...serializableRest,
+    selector,
+    source
+  } as unknown as T;
+};
```
oleksandr-danylchenko, line 156: Intentionally using the `as unknown as Something` because Typescript doesn't know that that `target` and `serializableRest` makes the expected type complete

packages/annotorious-core/src/model/Annotation.ts
```diff
@@ -12,16 +12,28 @@ export interface Annotation {
 
     [key: string]: any;
 
-  }
+  };
 
 }
 
-export interface AnnotationTarget {
+export interface AnnotationTarget extends AnnotationLifecycleInfo {
```
rsimon, line 19: Is it really necessary to introduce this as another concept? TBH looks a bit like overkill to me.

oleksandr-danylchenko: It was convenient to extract the properties of the lifecycle into another type for grouping and use it as a return value for the parse/serialize functions. Additionally, the types of the `AnnotationLifecycleInfo` & `W3CAnnotationLifecycleInfo` will be different, because the first one uses `Date` for the timestamps, and the second one uses `string`

oleksandr-danylchenko: Another way to group the lifecycle properties could be adding a type with a `Pick`. So it could look like this:
```ts
export interface AnnotationTarget {...}

type AnnotationLifecycleInfo = Pick<AnnotationTarget, 'creator' | 'created'>;
```
But it seems odd and inversed to me 🤔

oleksandr-danylchenko:
⏫ I'll be able to get back to this PR only in a ~week. The priorities on the project shifted and till the end of this week, I won't have enough time for annos. And from XMas I'm on vacation for a week.

rsimon:
Sure, no worries - and no hurry of course!

PR #345, opened 2024-01-04
Added parsing/serialization for the `update` props

oleksandr-danylchenko:
## Issue
This PR is related to the https://github.com/annotorious/annotorious/issues/325. As of now the `updated` & `updatedBy` properties are being destructed, but from annotation, but not processed. So then we'll be losing those props.

Was leaving those properties of the Core model a deliberate move? I thought that they belong there, as they are included in the `AnnotationTarget` 🤔

### DIFF
--- a/packages/annotorious/src/model/w3c/W3CImageFormatAdapter.ts
+++ b/packages/annotorious/src/model/w3c/W3CImageFormatAdapter.ts
@@ -33,8 +33,6 @@ export const parseW3CImageAnnotation = (
   const { 
     creator,
     created,
-    updatedBy,
-    updated,
     body, 
     ...rest 
   } = annotation;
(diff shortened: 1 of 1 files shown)


oleksandr-danylchenko:
Oh... I realized that there are no corresponding `updatedBy` and `updated` props in the W3C spec...
Instead, it only has the `modified` prop

oleksandr-danylchenko:
Oh... I realized that there are no corresponding `updatedBy` and `updated` props in the W3C spec...
Instead, it only has the [`modified` prop](https://www.w3.org/TR/annotation-model/#lifecycle-information)

oleksandr-danylchenko:
@rsimon, would it be appropriate to add parsing/serialization for the `modified` prop to not lose the `updated` info?
I also noticed that there are no `creator`/`created` props specified on the parent `W3CAnnotation` type too

oleksandr-danylchenko:
I also noticed a few smaller flaws:
- Missing `creator`/`created` props on the parent `W3CAnnotation` type
- Type of the `created` on `W3CAnnotationBody` is `Date` instead of `string`
- The `modified` prop from spec is not utilized. So the `updated` info can be lost

I think I'll address all that in a separate PR

PR #346, opened 2024-01-04
Added parsing/serialization of the `modified` prop from the W3C spec

oleksandr-danylchenko:
## Issue
This PR solves the https://github.com/annotorious/annotorious/issues/325

When working on the W3C adapter for the Text annotations I noticed a few possible mistakes in the parsing/serialization in the W3C adapter:
- Missing `creator`/`created` props on the parent `W3CAnnotation` type
- Type of the `created` on `W3CAnnotationBody` is `Date` instead of `string`
- The `modified` prop from the spec is not utilized. So the `updated` prop from the core model could be lost
- The `updated`/`updatedBy` was being destructured from the w3c annotation, but spec doesn't have such props


rsimon:
Many thanks! Merged & will be part of release 3.0.0-rc.17, which I'm going to publish later today.

PR #347, opened 2024-01-04
Added dedicated `W3CImageAnnotation` type definition

oleksandr-danylchenko:
## Issue
This PR resolves the https://github.com/recogito/text-annotator-js/issues/6 issue
When I was working on adding the  [`W3CTextAnnotationTarget`](https://github.com/recogito/text-annotator-js/pull/10/files#diff-ce397af0dfd03fe88dcf0fa9a95b7afd83f5bef452dffb6c5a83fa349e1710fbR9-R13) type I got blocked when I tried to extend it from the `W3CAnnotationTarget` with the next error: 
 
That was caused by the `W3CSelector` type that wasn't generic enough to handle selectors w/o the `value` prop, like `TextQuote`/`TextPosition` ones.

## Changes made
In this PR I replaced the `selector` type on the generic `W3CAnnotationTarget` with the `AbstractW3CSelector`. This is similar to how the selectors' variability is handled in the Core model: 
https://github.com/annotorious/annotorious/blob/f46d3a136befa698672c958cc918fd68a31f09f6/packages/annotorious-core/src/model/Annotation.ts#L19-L23
It helped me to declare the [`W3CTextAnnotationTarget`](https://github.com/recogito/text-annotator-js/pull/10/files#diff-ce397af0dfd03fe88dcf0fa9a95b7afd83f5bef452dffb6c5a83fa349e1710fbR9-R13) easily.

Nevertheless, to not lose types for the image annotations, I added a dedicated `W3CImageAnnotation` type with a specific `W3CImageSelector` type. That approach is unified with how the [`W3CTextAnnotation`](https://github.com/recogito/text-annotator-js/pull/10/files#diff-ce397af0dfd03fe88dcf0fa9a95b7afd83f5bef452dffb6c5a83fa349e1710fbR3-R7) declared


packages/annotorious/src/model/w3c/fragment/FragmentSelector.ts
```diff
@@ -1,16 +1,6 @@
...
-
-export interface FragmentSelector extends W3CSelector {
-
-  type: 'FragmentSelector';
-
-  conformsTo: 'http://www.w3.org/TR/media-frags/',
-
-  value: string;
-  
-}
```
oleksandr-danylchenko, line 13: It's debatable whether `FragmentSelector` and `SVGSelector` should have been moved to the `W3CImageAnnotation.ts`. But in my view, it's a bit easier to observe the whole type in one place. Plus it makes the structure similar to the one we'll have for the [`W3CTextAnnotation`](https://github.com/recogito/text-annotator-js/pull/10/files#diff-ce397af0dfd03fe88dcf0fa9a95b7afd83f5bef452dffb6c5a83fa349e1710fbR3-R7)

rsimon:
I agree that makes sense. Thanks for this - merged!

One minor thing: I moved the type declarations for `FragmentSelector` and `SvgSelector` back to their old locations. My convention (don't actually remember where I picked it up exactly) is to have source files with uppercase character only if there's a type of the same name defined in there. Also feels like declaration + associated parse/serialize utility functions are kept together this way.

Issue #365, opened 2024-02-06
Support for multiple `targets` attached to the same `Annotation`


oleksandr-danylchenko:
## Issue
This issue is a derivative of the https://github.com/recogito/text-annotator-js/issues/24 (Omission of the not annotable elements within the text). Its main goal is to have the ability to _split the selection into multiple ranges_ that will be transformed into [multiple targets](https://arc.net/l/quote/ktuowhmb) _attached to the same parent annotation_.

But, unfortunately, implementing the multiple `target`s support only on the `text-annotator-js` level would be quite a complicated task. Mainly because most of the generic types expect that the [`TextAnnotation`](https://github.com/recogito/text-annotator-js/blob/80c2f9235ef27b2e57ff84a23d1f100a7b377e83/packages/text-annotator/src/model/core/TextAnnotation.ts#L3) should directly extend the [`Annotation`](https://github.com/annotorious/annotorious/blob/8de692434213ad910df3af53fef28b620c60ffc9/packages/annotorious-core/src/model/Annotation.ts#L3-L17) type. This would be impossible if I'd wanna use `target: TextAnnotationTarget[]` type: ![incompatible types](https://github.com/annotorious/annotorious/assets/68850090/3e9e5d18-6644-404b-ae1a-2c6d93f5647a)

## Possible solutions
Replace the [`target`](https://github.com/annotorious/annotorious/blob/8de692434213ad910df3af53fef28b620c60ffc9/packages/annotorious-core/src/model/Annotation.ts#L7) property on the `Annotation` with the `targets: AnnotationTarget[]` following the patterns established for the [`bodies`](https://github.com/annotorious/annotorious/blob/8de692434213ad910df3af53fef28b620c60ffc9/packages/annotorious-core/src/model/Annotation.ts#L9).

Such a solution has a few major benefits to it:
1. _**Versatility. We're not limited to the number of targets in any of the derived types of annotations.**_
2. `bodies` has a pretty robust API already and can be an example of how the new `targets` property should be handled.
3. W3C standard compliance. We'll be able to process W3C annotations that have multiple targets assigned. [Which is officially allowed](https://www.w3.org/TR/annotation-model/#cardinality-of-bodies-and-targets)

oleksandr-danylchenko:
[...]

PR #367, opened 2024-02-07
#365 Added support for multiple `targets` in the `annotorious-core`

oleksandr-danylchenko:
## Issue
This PR addresses both issues #365 and https://github.com/recogito/text-annotator-js/issues/24. Where a need to store and process multiple targets attached to the same annotation was expressed.

## Changes Made
In this PR I replaced the singular `target` property with the _plural `targets: AnnotationTarget[]`_ and made it act basically _like the `bodies`_:
- each `target` has an `id` now
- `targetIndex` was added to the `Store` to bind `target⟡annotation`
- replaced the usage of `target` with the `targets` across the `annotorious-core`
- updated fixtures and tests to comply with the new format. They are passing 👌🏻 tests passing

## Further Steps
This PR covers ONLY the `annotorious-core` as the **_1st migration step_**. In the following PR the image annotation tool within the `annotorious` package will be updated


packages/annotorious-core/test/state/StoreObserver.fixtures.ts
oleksandr-danylchenko, line 1: For the fixtures, I did the next transforms:
1. Replaced `target` with `targets`, placing the value within an array
2. Replaced `targetUpdated` with `targetsUpdated` to match the format of the bodies

rsimon:
Sorry, you should have discussed this change before doing all that work. Have you considered changing the TextAnnotationTarget to simply support multiple Ranges? This would have been my recommendation, since it keeps the complexity where it's (currently) required, and doesn't drill down in to the image annotation core.

The PR introduces a hard breaking API change that breaks too much code in established codebases. (There are at least two large ones I'm working on myself, and another one I'm aware of.) Which means I will not merge this anytime soon. 

I realize you may consider multiple targets a valid scenario for images, too, and it may be, in the long run. But at the moment, there are too many open questions around it. Multi-target scenarios for images would impact the drawing tools architecture and UX; we'd need to sort out things like multi-select, grouping, etc. first. Even then I'd likely cover the use case with a single internal `target` that supports something like a `GeometryCollection` type or similar. (Keep in mind that the WebAnno notion of `target` is somewhat different. They do this, to some extent, to enable a graph model. An annotation could, for example, point to two different web pages and make a statement about the relation between them.)

Happy to discuss changes to the TextAnnotationTarget model. But the core API will have to stick to the single target I'm afraid.

oleksandr-danylchenko:
Welcome back! Thanks for looking through the PR 🙌🏻
First of all, I wanna specify that the main intentions of the PRs I made were to verify that our goal of having selections with "holes" is achievable using the `A9S` + `R6O/text-annotator-js` infrastructure. And fortunately - the answer is affirmative! Using multiple targets, I finally could make the canvas render the highlights only for the annotable elements.

image

So those PRs might not be merged as-is, but they show the direction we can take to achieve the "omission of the not annotable elements" goal.

---

> Have you considered changing the TextAnnotationTarget to simply support multiple Ranges?

Yep, I tried it first, but then I realized that it's [not a viable way](https://github.com/annotorious/annotorious/pull/367#issue-2122884751) to add support for multiple targets because:
1. It makes the text annotation fully incompatible with the parent `Annotation` type. That includes all the shared utilities that we use from the A9S package. 
2. It makes the `targets` different from the existing `bodies` property. However, they act in almost exactly similar ways when multiple targets can be associated with a single anno

> I realize you may consider multiple targets a valid scenario for images, too, and it may be, in the long run.

Nope. I'm not that particularly concerned about _using_ multiple targets for the images. The `target` -> `targets` update in A9S is done mainly to make the `text-annotator-js` compatible with the types and properties that are used by the "base" `Annotation` type.

rsimon:
My impression is that nothing prevents us from still having a single target (with multiple ranges) in the internal data model. Whether this is expressed as multiple targets in W3C doesn't really make a difference - that's what the crosswalk can handle. But internally, you could still stick with a single target. How the target is structured would be up to the text client, anyway. So as long as it knows how to handle them, there shouldn't be a need to push text-specifics into the image annotation client?

oleksandr-danylchenko:
Hmm... That's an interesting idea... 🤔 
Because changing the `selector` type on the `TextAnnotationTarget` shouldn't be disruptive for the parent `A9S`
image

Not sure why I didn't notice such an opportunity from the start, being too tied to the W3C's idea of mandatorily having multiple targets 🤷🏻‍♂️ 

---

Thanks! I'll give it a try!

oleksandr-danylchenko:

PR #368, opened 2024-02-07
#365 Added support for multiple `targets` in the `annotorious-core`

oleksandr-danylchenko:
## Relations
This PR is fully based on the https://github.com/annotorious/annotorious/pull/367 PR and includes all of its changes as well. So I would recommend starting from [here](https://github.com/annotorious/annotorious/pull/367) first

## Issue
This PR addresses the #365 issue where a need to store and process multiple targets attached to the same annotation was expressed.

## Changes made
In this PR I made the `ImageAnnotation` use the new `targets` property from the base `Annotation` class. Which also makes the sub-type compatible with all the shared helpers. 
Fortunately, the code within the A9S package almost didn't change. The most pragmatic solution was to just use `target[0]` instead of `target`. Because adding support for multiple selections on the image is a whole other hellish task.

---

For the testing, I used the existing attached app and I didn't notice any broken behavior after the change so far 🎉 

https://github.com/annotorious/annotorious/assets/68850090/8dc824fc-a4a8-42b3-b8d8-5edc02d26262

## Further Steps
This PR covers only the `annotorious` package as the **_2nd migration step_**. In the following PR, I'm about to address the react wrapper for the a9s


oleksandr-danylchenko:

PR #369, opened 2024-02-07
#365 Added support for multiple targets in the `annotorious-react` + `opeanseadragon`

oleksandr-danylchenko:
## Relations
This PR is fully based on the #367 and  https://github.com/annotorious/annotorious/pull/368 PRs. It includes all of their changes as well. So I would recommend starting from them first

## Issue
This PR addresses the #365 issue where a need to store and process multiple targets attached to the same annotation was expressed.

## Changes Mage
In this PR, following the #368, I replaced all the usages of the singular `target` with the `targets`. And the places where the target is being read, the `targets[0]` was enough to make the react+openseadragon examples work properly 👌🏻 

https://github.com/annotorious/annotorious/assets/68850090/008e52fa-f612-4155-8922-dbfb77ccf579

## Further Steps
This PR is the **_final 3rd migration step_** for the `target` -> `targets` within the [`annotorious`](https://github.com/annotorious/annotorious) package. And it opens the way to start working on the https://github.com/recogito/text-annotator-js/issues/24 in the `text-annotator-js` 🙌🏻


oleksandr-danylchenko:

PR #380, opened 2024-02-29
Added `replace` argument for the `setAnnotations`

oleksandr-danylchenko:
## Issue
`Annotator` type is a public API of the library:
https://github.com/annotorious/annotorious/blob/adc3f3b793de47731119efc1a35b8c03af2beb8c/packages/annotorious-core/src/model/Annotator.ts#L15-L65
The **_`setAnnotations`_** is a method to add annotations in a bulk way to the store using the provided `adapter`:
https://github.com/annotorious/annotorious/blob/adc3f3b793de47731119efc1a35b8c03af2beb8c/packages/annotorious-core/src/model/Annotator.ts#L148-L159
###### That function is a primary way to add multiple annotations at once w/o using the `loadAnnotations` method, which can be unsuitable in many scenarios (specific auth rules, offline-first env, specific app storage, etc.)

Unfortunately, on the subsequent runs of that function - it'll always replace all the existing annotations in the store 😱. This is not desired when the loading happens page by page and we don't want to load insert the same page time after time. 

## Changes Made
I added the `replace` argument that can be controlled from the outside of the library

### DIFF
--- a/packages/annotorious-core/src/model/Annotator.ts
+++ b/packages/annotorious-core/src/model/Annotator.ts
@@ -34,7 +34,7 @@ export interface Annotator<I extends Annotation = Annotation, E extends unknown 
 
   getUser(): User;
 
-  loadAnnotations(url: string): Promise<E[]>;
+  loadAnnotations(url: string, replace?: boolean): Promise<E[]>;
 
   redo(): void;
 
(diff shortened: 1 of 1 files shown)


packages/annotorious-core/src/model/Annotator.ts
```diff
@@ -145,16 +145,16 @@ export const createBaseAnnotator = <I extends Annotation, E extends unknown>(
     }
   }
 
-  const setAnnotations = (annotations: E[]) => {
+  const setAnnotations = (annotations: E[], replace = true) => {
```
oleksandr-danylchenko, line 148: `replace` is also `true` by default in the `bulkAddAnnotation`:
https://github.com/annotorious/annotorious/blob/adc3f3b793de47731119efc1a35b8c03af2beb8c/packages/annotorious-core/src/state/Store.ts#L136

rsimon:
Looks good, thanks!

Issue #382, opened 2024-03-07
`AnnotoriousPlugin` component doesn't reapply the `plugin` on its change


oleksandr-danylchenko:
## Required Behavior
When the plugin you want to apply to the annotator instance depends on dynamic arguments - it'll have the next design: 
```ts
const createWebtextsPlugin = (pageId) => (anno) => {
        ...
	const _addAnnotation = store.addAnnotation.bind(store);
	store.addAnnotation = (annotation, origin) => {
		const webtextsAnnotation = ...; // Some transforms that depend on the `pageId`
		return _addAnnotation(webtextsAnnotation, origin);
	};
        ...
} 
```
And on each change of the `pageId` the `createWebtextsPlugin will return the new function that should applied to the annotator instance:
```tsx
const webtextsPlugin = useMemo(() => createWebtextsPlugin(pageFamilyId), [pageFamilyId]);
```

## Issue

Unfortunately, such an approach won't work for the current A9S setup and you'll experience a few issues:
1. The `AnnotoriousPlugin` won't react to the `webtextsPlugin` changes❗
That's  because it omits the `plugin` from the `useEffect` deps list: https://github.com/annotorious/annotorious/blob/adc3f3b793de47731119efc1a35b8c03af2beb8c/packages/annotorious-react/src/AnnotoriousPlugin.tsx#L18-L27
So even if a user changes a page they looking at - annotations will still be transformed in the context of the previous one 🤷🏻‍♂️ 
2. Even after you add the `plugin` to the deps list - the monkey patching of the store will immediately backfire. And that's because on subsequent runs, the `const _addAnnotation = store.addAnnotation.bind(store);` will be the previously mutated `addAnnotation` method instance. So you'll experience a weird double processing of the annotation: ![image](https://github.com/annotorious/annotorious/assets/68850090/aa0cd48a-3097-46f4-a61d-233715855a26)

## Possible Solutions
1. Add the `plugin` to the deps list for the `AnnotoriousPlugin` to re-apply it on change
2. Specify in the examples that after mutating the store, it should be returned to the default state in the `unmount` callback:
```ts
export const createWebtextsPlugin =
	(pageId) =>
	(anno) => {
		const store = anno.state.store;

		const _addAnnotation = store.addAnnotation.bind(store);
		store.addAnnotation = (annotation, origin) => {
			...
			return _addAnnotation(webtextsAnnotation, origin);
		};

[...]

PR #383, opened 2024-03-07
#832 Added `plugin` to the deps list of the `AnnotoriousPlugin`

oleksandr-danylchenko:
## Issue
This PR addresses the #382 issue by making the `AnnotoriousPlugin` effect to re-apply the plugin on its change. Otherwise, the out-of-date plugin will still be applied to the store 🤷🏻‍♂️


oleksandr-danylchenko:
Re-applying the `plugin` that monkey-patches might lead to unexpected results: 
> On subsequent runs, the `const _addAnnotation = store.addAnnotation.bind(store);` will be the previously mutated `addAnnotation` method instance. So you'll experience a weird double processing of the annotation: image

PR #385, opened 2024-03-21
Added missing `@annotorious/openseadragon` to peer deps of `@annotorious/react`

oleksandr-danylchenko:
## Issue
The issue is similar to #384, so I would recommend checking it first.
The `@annotorious/openseadragon` is not being installed when you download the sources of the `@annotorious/react` and run `npm install`. And that leads to a failed `build` script 🤷🏻‍♂️

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5188,6 +5188,7 @@
       },
       "peerDependencies": {
         "@annotorious/annotorious": "*",
+        "@annotorious/openseadragon": "*",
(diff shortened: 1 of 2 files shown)


oleksandr-danylchenko:
Unfortunately, the `peerDependecies` are not installed automatically when using the `npm@10` 👀

rsimon:
Thanks!

PR #386, opened 2024-03-22
Made packages use standalone dependencies

oleksandr-danylchenko:
## Issues
When I tried using the packages as standalone units I noticed that each sub-package doesn't contain an exhaustive dependencies list that are required for its functioning and building. 

---

The dependencies tree for the packages looked like this:
| Package | Dependencies |
|--------|--------|
| `@annotorious/core` | ✅ ∅ | 

---

| Package | Dependencies |
|--------|--------|
| `@annotorious/annotorious` | ⚠️ `@annotorious/core` (wasn't specified as peer)  | 

---

| Package | Dependencies |
|--------|--------|
| `@annotorious/react` | ⚠️ `@annotorious/core` (wasn't specified as peer or dependency) | 
| | ⚠️ `@annotorious/annotorious` (wasn't specified as dependency) | 
| | ✅ `@annotorious/openseadragon` | 

---

| Package | Dependencies |
|--------|--------|
| `@annotorious/openseadragon` | ⚠️ `@annotorious/core` (wasn't specified as peer or dependency) | 
| | ⚠️ `@annotorious/annotorious` (wasn't specified as peer) | 

---

| Package | Dependencies |
|--------|--------|
| `@annotorious/react-manifold` | ⚠️ `@annotorious/react` (wasn't specified as dependency) | 

---

| Package | Dependencies |
|--------|--------|
| `@annotorious/svelte` | ⚠️ `@annotorious/core` (wasn't specified as peer or dependency) | 
| | ⚠️ `@annotorious/annotorious` (wasn't specified as dependency) | 
| | ⚠️ `@annotorious/openseadragon`  (wasn't specified as dependency) |

### DIFF
--- a/packages/annotorious-react-manifold/package.json
+++ b/packages/annotorious-react-manifold/package.json
@@ -39,7 +39,6 @@
     "vite-tsconfig-paths": "^4.3.1"
   },
   "peerDependencies": {
-    "@annotorious/react": "*",
     "openseadragon": "^3.0.0 || ^4.0.0",
     "react": "16.8.0 || >=17.x || >=18.x",
     "react-dom": "16.8.0 || >=17.x || >=18.x"
--- a/packages/annotorious-react-manifold/package.json
+++ b/packages/annotorious-react-manifold/package.json
@@ -43,5 +43,8 @@
     "openseadragon": "^3.0.0 || ^4.0.0",
     "react": "16.8.0 || >=17.x || >=18.x",
     "react-dom": "16.8.0 || >=17.x || >=18.x"
+  },
+  "dependencies": {
+    "@annotorious/react": "*"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -5214,6 +5214,9 @@
       "name": "@annotorious/react-manifold",
       "version": "3.0.0-rc.19",
       "license": "BSD-3-Clause",
+      "dependencies": {
+        "@annotorious/react": "*"
+      },
       "devDependencies": {
         "@types/react": "^18.2.48",
         "@types/react-dom": "^18.2.18",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5179,6 +5179,8 @@
       "version": "3.0.0-rc.19",
       "license": "BSD-3-Clause",
       "dependencies": {
+        "@annotorious/annotorious": "*",
+        "@annotorious/core": "*",
         "@annotorious/openseadragon": "*",
         "@neodrag/react": "^2.0.3"
       },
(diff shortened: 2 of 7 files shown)


packages/annotorious-react-manifold/package.json
```diff
@@ -39,9 +39,11 @@
     "vite-tsconfig-paths": "^4.3.1"
   },
   "peerDependencies": {
-    "@annotorious/react": "*",
     "openseadragon": "^3.0.0 || ^4.0.0",
     "react": "16.8.0 || >=17.x || >=18.x",
     "react-dom": "16.8.0 || >=17.x || >=18.x"
+  },
+  "dependencies": {
+    "@annotorious/react": "*"
```
oleksandr-danylchenko, line 47: Was moved to a `dependencies` as an essential dependency w/o which that package won't properly run or build. The "*" will also help to avoid version duplication in the consuming application

rsimon:
Thanks for this! In fact I noticed this recently when a client used the CJS compiled version. (It didn't really cause problems with ES & most bundlers. But with CJS and `<script>` import, things failed miserably without the dependencies.)

oleksandr-danylchenko:
Nice! Thanks for the merge 🔥 
When can we expect the package version bump? 🙏🏻 
It'll help us to us an actual library instead of our "re-published" A9S version

rsimon:
Almost... right now. I've updated all dependencies & now fighting a little with a PixiJS compat issue I need to resolve. Shouldn't take long though.

oleksandr-danylchenko:
> PixiJS compat issue I need to resolve

Oh! I saw it yesterday too! I thought that I messed something up locally with all the dependencies linking 🤯 
image

rsimon:
No, should be ok I think. In my case, there were some general TS errors in the Pixi v6 codebase. (They only surfaced after upgrading to a new TS version.) I managed to upgrade to PixiJS v7 now. New release is out. Thanks for the help!

https://github.com/annotorious/annotorious/releases/tag/v3.0.0-rc.20

oleksandr-danylchenko:
I think I spotted a bug in the usage of the `"depName": "*"` in `dependencies`.  As it implicitly _links_ the packages! 
If you try to pull a fresh repo w/o `dist` folders - you won't be able to build it because of the egg-chicken problem. To access the linked package - it needs to have a `dist`. And to build the linked package its dependencies also should have the `dist`. Then it cycles 🤯

oleksandr-danylchenko:
On my way to replace the wildcards with the exact versions

rsimon:
Argh... I feared that the wildcard would cause problems. But I was hoping to avoid it, because now I'd need to (manually) update the version numbers for each release. If you figure out a way to automate that, let me known. (Right now I'm using `npm version <new-version> --workspaces` which updates the package.json version numbers in sync. But no similar out-of-the-box solution seems to exist for updating anything in the dependencies list.

PR #387, opened 2024-03-22
Added exact package versions usage

oleksandr-danylchenko:
Fixes the wildcard bug that doesn't allow to build the project, see https://github.com/annotorious/annotorious/pull/386#issuecomment-2014975694. Specifically added usage of the `.19` version that doesn't have the "*"

### DIFF
--- a/packages/annotorious/package.json
+++ b/packages/annotorious/package.json
@@ -48,7 +48,7 @@
     "vitest": "^1.4.0"
   },
   "dependencies": {
-    "@annotorious/core": "*",
+    "@annotorious/core": "3.0.0-rc.19",
     "rbush": "^3.0.1",
     "uuid": "^9.0.1"
   }
--- a/packages/annotorious/package.json
+++ b/packages/annotorious/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@annotorious/annotorious",
-  "version": "3.0.0-rc.20",
+  "version": "3.0.0-rc.21",
   "description": "Add image annotation functionality to any web page with a few lines of JavaScript",
   "author": "Rainer Simon",
   "license": "BSD-3-Clause",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5131,8 +5131,7 @@
       }
     },
     "packages/annotorious": {
-      "name": "@annotorious/annotorious",
-      "version": "3.0.0-rc.20",
+      "version": "3.0.0-rc.21",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@annotorious/core": "3.0.0-rc.19",
@@ -5154,7 +5153,7 @@
     },
     "packages/annotorious-core": {
       "name": "@annotorious/core",
-      "version": "3.0.0-rc.20",
+      "version": "3.0.0-rc.21",
       "license": "BSD-3-Clause",
       "dependencies": {
         "dequal": "^2.0.3",
@@ -5177,7 +5176,7 @@
     },
     "packages/annotorious-openseadragon": {
       "name": "@annotorious/openseadragon",
-      "version": "3.0.0-rc.20",
+      "version": "3.0.0-rc.21",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@annotorious/annotorious": "3.0.0-rc.19",
@@ -5226,7 +5225,7 @@
     },
     "packages/annotorious-react": {
       "name": "@annotorious/react",
-      "version": "3.0.0-rc.20",
+      "version": "3.0.0-rc.21",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@annotorious/annotorious": "3.0.0-rc.19",
@@ -5256,7 +5255,7 @@
     },
     "packages/annotorious-react-manifold": {
       "name": "@annotorious/react-manifold",
-      "version": "3.0.0-rc.20",
+      "version": "3.0.0-rc.21",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@annotorious/react": "3.0.0-rc.19"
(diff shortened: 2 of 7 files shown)


oleksandr-danylchenko:
Yep, the `build` passed successfully now: 
image

oleksandr-danylchenko:
Oh, but using the `.19` version isn't be best option either, because it rolls back the pixi dependency in lock 🤔

oleksandr-danylchenko:
Oh...

oleksandr-danylchenko:
I think a better alternative would be to bump the versions across the packages and specify the new version everywhere instead of `.19` to avoid the unexpected rollback

rsimon:
Yes, that's exactly what I thought was needed. But, ideally, in a script supported way. (Not that I'm lazy or anything... but I know already I'll keep forgetting to update in some places every time a roll a new release :-D) For now, I'd say let's just fix it manually. But in the long run, some kind of automated solution would be great. I'm sure there must be something out there. Just hadn't had the time yet to do a thorough search.

oleksandr-danylchenko:
I suppose we can try using something like the [`bumpp`](https://www.npmjs.com/package/bumpp?activeTab=readme) that updates all the packages recursively. I'll try it out 👌🏻

oleksandr-danylchenko:
> I suppose we can try using something like the [`bumpp`](https://www.npmjs.com/package/bumpp?activeTab=readme) that updates all the packages recursively. I'll try it out 👌🏻

Nah, it doesn't update the `dependencies` out of the box ;(

oleksandr-danylchenko:
Seems like such a feature is supported only by `yarn` and its `yarn upgrade-interactive` command 🤷🏻‍♂️

oleksandr-danylchenko:
@rsimon, what is the usual deploy pattern here? Do you start from `a9s/core` and go up through the dependencies list?
1. @annotorious/core
2. @annotorious/annotorious
3. @annotorious/openseadragon
4. @annotorious/react
5. @annotorious/react-manifold
6. @annotorious/svelte

oleksandr-danylchenko:
Could I please get the `3.0.0-rc.21` deployed? 🙏🏻 🙏🏻🙏🏻
I'll unblock me with migration to the `@annotorious` dependencies in my project instead of littering NPM with "proxy" packages

rsimon:
> Could I please get the 3.0.0-rc.21 deployed? 🙏🏻 🙏🏻🙏🏻
I'll unblock me with migration to the @annotorious dependencies in my project instead of littering NPM with "proxy" packages

The current latest state of the repo is published on npm as 3.0.0-rc.20 (from just an hour ago or so)

oleksandr-danylchenko:
Yeah, but `.20` uses the `"*"` dependencies internally **_that don't allow building the packages_**. So the new version, where all the dependencies will point to the `^3.0.0-rc.21`, is needed

oleksandr-danylchenko:
Also having the new version will unblock this PR - https://github.com/recogito/text-annotator-js/pull/75

packages/annotorious/package.json
```diff
@@ -48,7 +48,7 @@
     "vitest": "^1.4.0"
   },
   "dependencies": {
-    "@annotorious/core": "*",
+    "@annotorious/core": "3.0.0-rc.19",
```
oleksandr-danylchenko, line 51: *Can only be bumped to the `^3.0.0-rc.21` after the respective version will get published on NPM

oleksandr-danylchenko:
Please 🙏🏻 
It's the only think that blocks me from moving forward through my workday

rsimon:
Ok, just to confirm. That would mean whenever a new version gets published (say, 3.0.0-rc.21) the dependencies could only ever point to the previous version, right? Hm, not ideal. But a) there's probably no way around that; b) when people use the package, a `^3.0.0-rc.20` would probably cause npm to pull the latest version (3.0.0-rc.21) anyway.

Re build order: I'm using npm workspaces, which defines it [in the monorepo package.json](https://github.com/annotorious/annotorious/blob/main/package.json). The order is as you say:

* annotorious-core
* annotorious
* annotorious-openseadragon
* annotorious-react
* annotorious-react-manifold
* annotorious-svelte

rsimon:
I can publish the new release now. But I'll bump the dependency versions to ^3.0.0-rc.20, ok? I.e. I'll close this PR & make that change.

oleksandr-danylchenko:
> That would mean whenever a new version gets published (say, 3.0.0-rc.21) the dependencies could only ever point to the previous version, right

Not exactly... The process to make them all use the same versions can look like this:
1. Bump all the versions to `.21`
2. Publish `@a9s/core` to NPM
3. Install `@a9s/core@.21` instead of `.19` in the packages
4. Repeat steps 2-3 in the workspaces order

In that way, all the packages will have the `.21` version and will depend on `.21` versions too

oleksandr-danylchenko:
> b) when people use the package, a ^3.0.0-rc.20 would probably cause npm to pull the latest version (3.0.0-rc.21) anyway.

Also correct. Unless their package-lock has captured the `.20` version

rsimon:
Yes, but that means you'd have to publish each package individually, and then, after publishing one package, update the version numbers in the next, publish the next, update version number in the next etc. Again, possible to script, but not my favourite workflow I guess.

oleksandr-danylchenko:
> Yes, but that means you'd have to publish each package individually, and then, after publishing one package, update the version numbers in the next, publish the next,

Unfortunately, having the current setup, it's the only way to make everything use the same version 🤷🏻‍♂️

rsimon:
v3.0.0-rc.21 is up!

oleksandr-danylchenko:
Thanks!

PR #388, opened 2024-03-25
Excluded mandatory dependencies from bundling for ES modules

oleksandr-danylchenko:
## Issue
When working on the consuming app for the `A9S` + `R6O` packages I noticed that the libraries code is [included in the final bundle twice](https://github.com/recogito/text-annotator-js/pull/75#issuecomment-2018235214) 👁️👁️ 
I found it to be an error-prone approach, especially if the React Context gets duplicated and the hooks lose a "single source of truth" instance.

## Changes Made
I excluded the mandatory dependencies from the bundling of the packages. It gives us a few benefits:
1. The dependencies are guaranteed to be installed by the consuming app/package. Therefore, the A9S should expect that those dependencies will be available to consume
2. There will be no duplication of the code in the final bundle of the consuming app/package
3. Smaller package size to publish

## Next Steps
Apply the same treatment of the R6O dependencies

### DIFF
--- a/packages/annotorious-openseadragon/vite.config.js
+++ b/packages/annotorious-openseadragon/vite.config.js
@@ -8,11 +8,7 @@ import * as packageJson from './package.json';
 export default defineConfig({
   plugins: [
     svelte({ preprocess: sveltePreprocess() }),
-    dts({
-      insertTypesEntry: true,
-      include: ['./src/'],
-      entryRoot: './src'
-    })
+    dts({ insertTypesEntry: true, include: ['./src/'], entryRoot: './src' })
   ],
   server: {
     open: '/test/index.html'
@@ -28,7 +24,8 @@ export default defineConfig({
     },
     rollupOptions: {
       external: [
-        ...Object.keys(packageJson.peerDependencies)
+        ...Object.keys(packageJson.peerDependencies),
+        /@annotorious\/(core|annotorious)/
       ],
       output: {
         assetFileNames: 'annotorious-openseadragon.[ext]',
--- a/packages/annotorious-openseadragon/vite.config.js
+++ b/packages/annotorious-openseadragon/vite.config.js
@@ -25,7 +25,8 @@ export default defineConfig({
     rollupOptions: {
       external: [
         ...Object.keys(packageJson.peerDependencies),
-        /@annotorious\/(core|annotorious)/
+        '@annotorious/core',
+        '@annotorious/annotorious'
       ],
       output: {
         assetFileNames: 'annotorious-openseadragon.[ext]',
--- a/packages/annotorious-openseadragon/vite.config.js
+++ b/packages/annotorious-openseadragon/vite.config.js
@@ -23,16 +23,10 @@ export default defineConfig({
         format === 'umd' ? `annotorious-openseadragon.js` : `annotorious-openseadragon.es.js` 
     },
     rollupOptions: {
-      external: [
-        ...Object.keys(packageJson.peerDependencies),
-        '@annotorious/core',
-        '@annotorious/annotorious'
-      ],
+      external: Object.keys(packageJson.peerDependencies),
       output: {
         assetFileNames: 'annotorious-openseadragon.[ext]',
         globals: {
-          '@annotorious/core': 'AnnotoriousCore',
-          '@annotorious/annotorious': 'Annotorious',
           openseadragon: 'OpenSeadragon'
         }
       }
--- a/packages/annotorious-openseadragon/vite.config.js
+++ b/packages/annotorious-openseadragon/vite.config.js
@@ -30,6 +30,7 @@ export default defineConfig({
       output: {
         assetFileNames: 'annotorious-openseadragon.[ext]',
         globals: {
+          '@annotorious/core': 'AnnotoriousCore',
           '@annotorious/annotorious': 'Annotorious',
           openseadragon: 'OpenSeadragon'
         }
--- a/packages/annotorious-react-manifold/vite.config.js
+++ b/packages/annotorious-react-manifold/vite.config.js
@@ -29,10 +29,7 @@ export default defineConfig({
       ],
       output: {
         preserveModules: true,
-        assetFileNames: 'annotorious-react-manifold.[ext]',
-        globals: {
-          react: 'React'
-        }
+        assetFileNames: 'annotorious-react-manifold.[ext]'
       }
     },
     sourcemap: true
--- a/packages/annotorious-svelte/vite.config.js
+++ b/packages/annotorious-svelte/vite.config.js
@@ -24,11 +24,15 @@ export default defineConfig({
       fileName: (format) => `annotorious-svelte.${format}.js`
     },
     rollupOptions: {
-      external: [...Object.keys(packageJson.peerDependencies)],
+      external: [
+        ...Object.keys(packageJson.peerDependencies),
+        /@annotorious\/(core|annotorious|openseadragon)/
+      ],
       output: {
         preserveModules: true,
         assetFileNames: 'annotorious-svelte.[ext]',
         globals: {
+          '@annotorious/annotorious': 'Annotorious',
           openseadragon: 'OpenSeadragon'
         }
       }
--- a/packages/annotorious-svelte/vite.config.js
+++ b/packages/annotorious-svelte/vite.config.js
@@ -26,7 +26,9 @@ export default defineConfig({
     rollupOptions: {
       external: [
         ...Object.keys(packageJson.peerDependencies),
-        /@annotorious\/(core|annotorious|openseadragon)/
+        '@annotorious/core',
+        '@annotorious/annotorious',
+        '@annotorious/openseadragon'
       ],
       output: {
         preserveModules: true,
--- a/packages/annotorious-svelte/vite.config.js
+++ b/packages/annotorious-svelte/vite.config.js
@@ -30,11 +30,7 @@ export default defineConfig({
       ],
       output: {
         preserveModules: true,
-        assetFileNames: 'annotorious-svelte.[ext]',
-        globals: {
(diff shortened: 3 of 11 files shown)


packages/annotorious-openseadragon/vite.config.js
```diff
@@ -28,15 +24,18 @@ export default defineConfig({
     },
     rollupOptions: {
       external: [
-        ...Object.keys(packageJson.peerDependencies)
+        ...Object.keys(packageJson.peerDependencies),
+        '@annotorious/core',
+        '@annotorious/annotorious'
```
oleksandr-danylchenko: I'm using the explicit exclusion list definition, because `/@annotorious\/(core|annotorious)/` leads to the following warning on bundling:
image

packages/annotorious-openseadragon/vite.config.js
```diff
@@ -28,15 +24,18 @@ export default defineConfig({
...
       ],
       output: {
         assetFileNames: 'annotorious-openseadragon.[ext]',
         globals: {
+          '@annotorious/core': 'AnnotoriousCore',
           '@annotorious/annotorious': 'Annotorious',
```
oleksandr-danylchenko: @rsimon, could you provide a bit more info about the `umd` consumers and where the global `Annotorious` variable is set? Is it possible to add the `AnnotoriousCore` there as well?

packages/annotorious-react-manifold/vite.config.js
```diff
@@ -29,12 +29,9 @@ export default defineConfig({
       ],
       output: {
         preserveModules: true,
-        assetFileNames: 'annotorious-react-manifold.[ext]',
-        globals: {
-          react: 'React'
-        }
```
oleksandr-danylchenko, line 35: `globals` definition is unused when only the `es` format is specified

packages/annotorious-openseadragon/vite.config.js
oleksandr-danylchenko: > Is it possible to add the AnnotoriousCore there as well?

If not, I think it's not a huge deal.  Bundling the core at least doesn't have global instances that can be duplicated or misused in the consuming apps/packages

packages/annotorious-openseadragon/vite.config.js
rsimon: But wouldn't that mean core and annotorious will not get bundled into the UMD bundle? That would IMO lead back to the broken script imports I had a while back.

packages/annotorious-openseadragon/vite.config.js
rsimon: I think this might not be needed and/or a mistake by me. If the everything gets bundled cleanly into the main bundle (=global AnnotoriousOSD), then this can be skipped. OpenSeadragon should the only necessary global.

packages/annotorious-svelte/vite.config.js
```diff
@@ -24,15 +24,17 @@ export default defineConfig({
       fileName: (format) => `annotorious-svelte.${format}.js`
     },
     rollupOptions: {
-      external: [...Object.keys(packageJson.peerDependencies)],
+      external: [
+        ...Object.keys(packageJson.peerDependencies),
+        '@annotorious/core',
+        '@annotorious/annotorious',
+        '@annotorious/openseadragon'
+      ],
```
rsimon, line 32: I think in this case, this might be ok. Because we're only publishing an ES package here?

packages/annotorious/vite.config.js
```diff
@@ -21,14 +17,20 @@ export default defineConfig({
...
-      fileName: (format) => 
-        format === 'umd' ? `annotorious.js` : `annotorious.es.js` 
+      fileName: (format) =>
+        format === 'umd' ? `annotorious.js` : `annotorious.es.js`
     },
     rollupOptions: {
+      external: [
+        '@annotorious/core',
+        'test/*'
+      ],
```
rsimon: But here, again, I think this would break the UMD bundle.

rsimon (changes requested):

packages/annotorious-openseadragon/vite.config.js
oleksandr-danylchenko: > But wouldn't that mean core and annotorious will not get bundled into the UMD bundle? 

That's correct. Their usage in the bundle will be replaced with references to the global variables `AnnotoriousCore` and `Annotorious` respectively. So the consuming app/package becomes responsible for setting those global variables

oleksandr-danylchenko: It's worth mentioning that the _`annotorious/annotorious` has already been excluded from the bundle_ before the https://github.com/annotorious/annotorious/pull/386

packages/annotorious-svelte/vite.config.js
oleksandr-danylchenko: Exactly! Reference to the code from those packages will be left untouched and we'll just have a bunch of `import ... from '@annotorious/...'` at the top of the resulting bundle

packages/annotorious/vite.config.js
oleksandr-danylchenko: _It might... 🤔_ 
That's why I [asked](https://github.com/annotorious/annotorious/pull/388/files/ee552bc3a77ed5ef30e9fb631a509b0773593196#r1537904480) about the places where/whether the variables for the `globals` definitions are set in the consuming apps/packages. 
For example, previously the `annotorious/annotorious` was already excluded from the bundling, being part of the `peerDependecies`. But it was being replaced with the `Annotorious` global variable. So, I suppose, this variable must have been set somewhere down the line 🤔

oleksandr-danylchenko: *I'm dying from the italicized emojis _🤣_ 
image

oleksandr-danylchenko: From [this answer](https://github.com/annotorious/annotorious/pull/388/files/ee552bc3a77ed5ef30e9fb631a509b0773593196#r1539111493), I surmise that the `Annotorious` variable was never set 🤷🏻‍♂️ 

**_So we might need to bundle the dependencies for the UDM compliance_**

---

However, if the package was already excluded from the bundle, the variable wasn't set and there were no complaints... Maybe no-one tried to use it w/o a bundler that uses es modules 😅

rsimon: Not sure I understand the question. If you build the UMD library, Vite will create a global `Annotorious` variable, with all the properties you would otherwise import as named imports. I.e. the following code with the ES bundle:

```js
import { createImageAnnotator } from '@annotorious/annotorious';

const anno = createImageAnnotator(...);
```

will become this...

```js
var anno = Annotorious.createImageAnnotator(...);
```

...when you use the script import. Therefore, if @annotorious/core is not bundled into the UMD module as a direct depenendency, the UMD bundle will fail. Same for the AnnotoriousOSD UMD bundle, which needs to have both `@annotorious/core` and `@annotorious/annotorious` bundled in to work.

rsimon: > So we might need to bundle the dependencies for the UDM compliance

Yes, I think we are talking about the same thing :-)

oleksandr-danylchenko:
I'll also check whether it's possible to do a bundling exclusion based on the formats 🤔.
As using the suggested setup should be fine and safe for all the ES consumers. But the UDM consumers can indeed lack the `annotorious/core` and/or `annotorious/annotorious`.
So, ideally, the resulting ES file would be pretty lean, but the UMD one can be more extended

oleksandr-danylchenko:
> So, ideally, the resulting ES file would be pretty lean, but the UMD one can be more extended

Unfortunately, Rollup doesn't allow such a setup for the external dependencies 😢. So we're obligated to bundle them both into ES & UMD modules.

---

Or... **The consumers can just include the external dependency for the `@annotorious/core` by themselves!** 💡 
The change from us would be to bundle the `@annotorious/core` into the UMD format under the `AnnotoriousCore` global variable. And then, [in the docs](https://github.com/annotorious/annotorious/blob/90d7b559fb7f588b6ac01fa09d2a4a9453d335c3/README.md?plain=1#L45-L48), add one more `script` link:
```diff
+<script src="https://cdn.jsdelivr.net/npm/@annotorious/core@latest/dist/annotorious-core.js"></script>
<script src="https://cdn.jsdelivr.net/npm/@annotorious/annotorious@latest/dist/annotorious.js"></script>
```

IMHO, such a solution might be a _better alternative in the long run_. As there will be less code duplication in the resulting bundles and less risk for [weird bugs](https://github.com/recogito/text-annotator-js/pull/75#issuecomment-2018235214) caused by it 🤔

rsimon:
Not a good dev experience though. I definitely want to keep it a single script import.

oleksandr-danylchenko:
Agree to disagree then 🤝 
My role was to suggest the option. So I'll add back bundling of the `/core` and `/annotorious` dependencies for the packages that have the `umd` as the target format

oleksandr-danylchenko:
*applied the same treatment to the https://github.com/recogito/text-annotator-js/pull/75 as well

rsimon:
> My role was to suggest the option. 

Definitely. And no offense taken, of course. Sorry if my reply might have come across a bit harsh (or, at least, very short ;-) I believe being able to keep things simple is worth the extra 26kB for now. (At least for Annotorious v2, the number of people using the script import still outweighs the bundler-based projects by far.)

oleksandr-danylchenko:
> Sorry if my reply might have come across a bit harsh

No, not at all! I just wanted to express that I left a suggestion, and you, as the author of the lib, _are in full right to push back on it or follow a specific preferred pattern_. 
I guess my phrasing might sounded a bit harsh or rude due to English being far from my 1st language, sorry 😅

oleksandr-danylchenko:
> Agree to disagree then 🤝
> My role was to suggest the option.

image

PR #402, opened 2024-05-14
Added baseline `Sorter` helper definition

oleksandr-danylchenko:
#### See - https://github.com/recogito/text-annotator-js/issues/94

In this PR I added a baseline type definition and implementation for the `Sorter` helper. 
It follows the `Filter` definition structure.

### DIFF
--- a/packages/annotorious-core/src/model/Sorter.ts
+++ b/packages/annotorious-core/src/model/Sorter.ts
@@ -1,0 +1,3 @@
+import type { Annotation } from './Annotation';
+
+export type Sorter<T extends Annotation = Annotation> = (a: T, b: T) => number;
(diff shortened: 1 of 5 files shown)


oleksandr-danylchenko:
I wasn't sure whether the `Sorter` should exactly follow the `Filter` steps. As I don't see an immediate application for it in the A9S suite 🤔. However, I suppose that the ordered overlapping might still be a valid concern even here.

packages/annotorious-core/src/model/Sorter.ts
```diff
@@ -0,0 +1,10 @@
+import type { Annotation } from './Annotation';
+
+export type Sorter<T extends Annotation = Annotation> = (a: T, b: T) => number;
+
+export const DEFAULT_ANNOTATION_SORTER: Sorter = (a, b) => {
+  const { target: { created: createdA } } = a;
+  const { target: { created: createdB } } = b;
+
+  return (createdA?.getTime() || 0) - (createdB?.getTime() || 0);
+};
```
oleksandr-danylchenko, line 10: Also, I'm not quite sure that this is an appropriate place to put the default implementation for the sorting function. But I think it's decently clear 🤷🏻‍♂️

PR #403, opened 2024-05-16
Added `modified` field handling for the `W3CAnnotationBody` parsing/serialization

oleksandr-danylchenko:
## Issue
Only the `created` lifecycle field was parsed/serialized for the `W3CAnnotationBody`, but not the [`modified`](https://arc.net/l/quote/yrqljguv). 

Also, including the raw `updated: Date` prop into the `W3CAnnotationBody` payload can lead to serialization warnings in such tools as Redux: 
image

## Changes Made
I followed the `modified` parsing/serialization pattern [that's already used](https://github.com/annotorious/annotorious/blob/98a9fa6011c3998fb46912fe6d68bd8af4b8de89/packages/annotorious/src/model/w3c/W3CImageFormatAdapter.ts#L64) for the `W3CImageAnnotation`. That includes mapping between `updated` ↔️ `modified` and exclusion of the `updated` from the w3c payload


rsimon:
Looks good, thanks!

PR #409, opened 2024-06-14
Fixed missing observers cleanup in the `.off` method

oleksandr-danylchenko:
## Issue - https://github.com/recogito/text-annotator-js/issues/107#issue-2351556561


rsimon:
Phew - thanks so much for discovering & debugging this! (And thanks for fixing the typo, too ;-)

packages/annotorious-core/src/lifecycle/Lifecycle.ts
```diff
@@ -35,8 +35,8 @@ export const createLifecyleObserver = <I extends Annotation, E extends unknown>(
     const callbacks = observers.get(event);
     if (callbacks) {
       const idx = callbacks.indexOf(callback);
-      if (idx > 0)
-        callbacks.splice(callbacks.indexOf(callback), 1);
+      if (idx !== -1)
+        callbacks.splice(idx, 1);
```
rsimon, line 39: I'm really not sure what I must have smoked that day :-D

oleksandr-danylchenko: I got shot myself with the same exact comparison more times than I would like to admit 😆

Issue #417, opened 2024-07-04
?Unexpected? `pointerEvent` removal from the `selection` upon store observation


oleksandr-danylchenko:
## Issue
In the A9S React wrapper, we observe the selection both from the `AnnotoriousContext['selection']` : 
https://github.com/annotorious/annotorious/blob/70704252789245ef17100d88ff8f54f3c48c328b/packages/annotorious-react/src/Annotorious.tsx#L67-L74
and the `store` updates:
https://github.com/annotorious/annotorious/blob/70704252789245ef17100d88ff8f54f3c48c328b/packages/annotorious-react/src/Annotorious.tsx#L76-L87
In the latter, `setSelection` updates only the `selection` prop, which dismisses the `pointerEvent` and makes in undefined.

##### (in the `text-annotator-js`)

That effect can be observed when you start selecting using the mouse and then select further using keyboard (`Shift + ←/→`):

https://github.com/annotorious/annotorious/assets/68850090/a1863f83-8ff6-4081-ac71-5b39b4847fd3

Or even when you select a longer text:

https://github.com/annotorious/annotorious/assets/68850090/5f831b23-38dd-43a1-9961-e58e8495ca7e

PR #418, opened 2024-07-04
#417 Fixed losing of the `pointerevent` during the selection

oleksandr-danylchenko:
## Changes Made
Added spreading for the selection state before updating the `selected`


rsimon:
Thanks, I think that change makes sense.

Issue #419, opened 2024-07-08
Generalize the `SelectionState` type to allow custom selection implementations


oleksandr-danylchenko:
## Context
Currently, the `SelectionState` and the `Selection` modules, in general, are designed to work exclusively with pointer-based user interactions. https://github.com/annotorious/annotorious/blob/9188ea1d73911209dcd367734e7f45f5382950e3/packages/annotorious-core/src/state/Selection.ts#L5-L10

## Issue
This is acceptable for the visual images editor, but it's limiting for the descending `@recogito/text-annotator-js`. The latter should support both the mouse and the keyboard selection simultaneously. 

When I tried adding a custom `TextSelection` module and its specific type with the `KeyboardEvent` support I stumbled into _type mismatches_ with the rest of the parent A9S infrastructure.

That's because the parent `AnnotatorState` is tightly coupled with the pointer-based `SelectionState`: https://github.com/annotorious/annotorious/blob/9188ea1d73911209dcd367734e7f45f5382950e3/packages/annotorious-core/src/model/Annotator.ts#L69-L79
And I couldn't pass my custom text selection into other A9S methods like `createLifecycleObserver`: 
https://github.com/annotorious/annotorious/blob/9188ea1d73911209dcd367734e7f45f5382950e3/packages/annotorious-core/src/lifecycle/Lifecycle.ts#L9-L10

## Suggested Changes
The `Selection` module in the `@a9s/core` may become a generic container and API requirements for the concrete selection implementations. The latter will be provided by the `ImageAnnotator`, or the `TextAnnotator`. 

That would make it similar to the [`baseRenderer`](https://github.com/recogito/text-annotator-js/blob/main/packages/text-annotator/src/highlight/baseRenderer.ts) we use in `text-annotator-js`.

The current `createSelectionState` implementation may be moved under the `ImageAnnotatorSelectionState.ts`.

rsimon:
Would it be sufficient to make the event a basic `Event` type? Or type the SelectionState with `<T extends Event>`? 

In general: I'd love to hear more about your use case. So far, I really only used the `pointerEvent` to get a screen coordinate for placing an editor popup. I guess a keyboard event wouldn't have such an option? Also, I'd be generally interested to improve the accessibility of both Annotorious and the Text Annotator, but don't really know how keyboard selection would work in either case. (For text, I assume, you still start with a pointer event, and then use `SHIFT` + arrow key to extend the selection?)

oleksandr-danylchenko:
[...]

PR #420, opened 2024-07-08
#419 Selection `event` generalization

oleksandr-danylchenko:
## Changes Made
1. Replaced the `pointerEvent` with the generic `event` that can be triggered either by the `Keyboard` or the `Pointer`.
2. Replaced `clickSelect` with `userSelect` to signify that the selection is triggered by the user-induced event, and not be tied to the "click" concept
3. Renamed `PointerSelectAction` -> `UserSelectAction`

### DIFF
--- a/docs/getting-started-image.md
+++ b/docs/getting-started-image.md
@@ -61,15 +61,15 @@ anno.on('createAnnotation', function(annotation) {
 
 ## AnnotoriousOpts
 
-| Arg                   | Type                                            | Default       |
-|-----------------------|-------------------------------------------------|---------------|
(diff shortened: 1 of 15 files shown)


rsimon:
Thanks for this! Looking good so far. One thing though: I'm not quite happy with the renaming. The original reason for naming it `pointerSelectAction` (rather than `selectAction`) was to differentiate it from programmatic selection through the application. So I guess we might call it `userSelectAction` or something?

oleksandr-danylchenko:
> So I guess we might call it userSelectAction or something?

Thanks, I like that suggestion!

Issue #377, opened 2024-02-26
Rename the `AnnotoriousPlugin` cleanup callback `unmount` -> `destroy` 


oleksandr-danylchenko:
## Current State
Currently, the `AnnotoriousPluginProps` receives the `plugin` function argument that returns the `{ unmount: ... }` object:
https://github.com/annotorious/annotorious/blob/adc3f3b793de47731119efc1a35b8c03af2beb8c/packages/annotorious-react/src/AnnotoriousPlugin.tsx#L5-L11

## Issue
Could we rename the `unmount` to `destroy`? Seems like it's a more popular name for the cleanup within the library and is used for most of the modules: 
https://github.com/annotorious/annotorious/blob/adc3f3b793de47731119efc1a35b8c03af2beb8c/packages/annotorious/src/Annotorious.ts#L107-L118

oleksandr-danylchenko:
> Plugins will typically have a `mountPlugin` method (which is used in vanilla JS). I'd like to keep `unmount` as the inverse name.

https://github.com/annotorious/annotorious/pull/427#issuecomment-2255664549

PR #427, opened 2024-07-25
Renamed `unmount` -> `destroy` in the `AnnotatorPlugin`

oleksandr-danylchenko:
## Issue - https://github.com/annotorious/annotorious/issues/377


rsimon:
Plugins will typically have a `mountPlugin` method (which is used in vanilla JS). I'd like to keep `unmount` as the inverse name.

oleksandr-danylchenko:
> Plugins will typically have a mountPlugin method

Interesting, I didn't know that... 🤔

PR #430, opened 2024-07-27
Added annotation-oriented convenience React hooks

oleksandr-danylchenko:
## Issue
1. There was no library-provided React hook for tracking changes for individual annotation. You could use smth like:
```tsx
const store = useAnnotationStore();
const annotation = store?.getAnnotation(id);
```
But that captures only a single "snapshot" of the annotation when the `getAnnotation` runs. It won't react to the subsequent underlying updates.
2. There was no simple React-based way to track the selection action for individual annotation by the id.

## Changes Made
1. Added the `useAnnotation` hook that reads the `store` for the 1st time and subscribes to the store update made for that annotation. It's convenient for the consuming app's components that want to know the latest `target`/`body` values and react to them accordingly.
2. Added the `useAnnotationSelectAction`. It's convenient for the consumer components that want to update the look based on the `UserSelectAction` for the annotations that are not yet selected. For example - the side note component should become "inert" and disabled when the `UserSelectAction` is `NONE`.

### DIFF
--- a/packages/annotorious-core/src/model/Annotator.ts
+++ b/packages/annotorious-core/src/model/Annotator.ts
@@ -1,9 +1,17 @@
 import type { Annotation } from './Annotation';
 import type { User } from './User';
-import type { PresenceProvider } from '../presence/PresenceProvider';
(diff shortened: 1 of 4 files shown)


rsimon:
Looks useful, thanks!

PR #432, opened 2024-08-01
Applying style based on props

po8rewq:
When you use the `style` prop on `ImageAnnotator`, the style isn't applied properly

Example of style:

```
{
      fill: "#FFFFFF",
      stroke: "#FF0000",
      strokeOpacity: 0.4,
 };
```

Before:
image

After:
image


rsimon:
Great,  thanks for this fix!

PR #433, opened 2024-08-02
Exposed `userSelectAction` prop from the `Selection`

oleksandr-danylchenko:
## Issue
We pass and persist the `userSelectAction` within the `Selection` module. But, unfortunately, there's no way to obtain its value from the annotator instance. The `state.selection` doesn't expose it 🤷🏻‍♂️.  

Therefore, it requires duplicating the `userSelectAction` in the consuming apps to make accessing it more convenient from nested components. Instead, it would be simpler to read it like:
```tsx
const r = useAnnotator();
console.log(r.state.selection.userSelectAction)
```


oleksandr-danylchenko:
The only issue is that the `userSelectAction` value is not "reactive". The `useEffect`/`useMemo` won't react to its update via the `setUserSelectAction` and will treat it as an opaque mutation.

The same issue is relevant to many other properties within the `text-annotator-js`, like the `annotatitingEnabled`, `filter`, `style`, etc. They all are persisted within the non-reactive JS modules.

However, I'm not sure what "quick&easy` solution can be applied here 🤷🏻‍♂️. Because migrating to the "reactive" store like Zustand or rewriting the state with the Svelte Stores will require quite a bit of effort and QC...

rsimon:
I don't know about this. It feels like kind of a weird pattern to me, in vanilla JS. In many cases, the user select action will be a function. There's probably not much you can do with it, except compare it with some locally held copy.

Arguably, this makes a lot more sense in React, where you're naturally dealing with props. I need to think about it. But wondering if that's not something I'd want to keep as is in the vanilla version, and only cover (through a state) in the React package specifically?

oleksandr-danylchenko:
> There's probably not much you can do with it, except compare it with some locally held copy

Yep, the issue with not exposing it, is that you need to preserve a local copy somewhere.

> and only cover (through a state) in the React package specifically?

Yes, it plays a bigger role there. However, I don't think that the vanilla API would "suffer" from providing that prop to the consumers, IMHO. Then the React user will be able to get it as well. So I would leave it in the vanilla version too.
The only thing that would make a significant React-specific improvement is making that property "reactive". But that's a bigger story that would take more effort I suppose.

rsimon:
If I see correctly, the initial `UserSelectActionExpression` is no longer passed to `createSelectionState`. So this would ignore any `UserSelectActionExpression` from the Annotorious init options?

oleksandr-danylchenko:
> the initial UserSelectActionExpression is no longer passed to createSelectionState

Yep, it's no longer passed to the constructor of the selection state.

> So this would ignore any UserSelectActionExpression from the Annotorious init options?

However, immediately after instantiation, the `setUserSelectAction` is called:
```ts
const selection = createSelectionState(store);
selection.setUserSelectAction(opts.userSelectAction);
```
Therefore, the initial `userSelectAction` value is still being respected and passed to the selection state.

---

The same approach is already used for the `SelectionHandler` module in `r6o` text annotator ([source](https://github.com/recogito/text-annotator-js/blob/21362cda552e8d2047af9dd3a6dadc4faee05ae2/packages/text-annotator/src/TextAnnotator.ts#L73-L74)):
```ts
const selectionHandler = SelectionHandler(container, state, opts.annotatingEnabled, opts.offsetReferenceSelector);
selectionHandler.setUser(currentUser);
```

PR #437, opened 2024-08-20
Added comparison before clearing the selection

oleksandr-danylchenko:
## Context
When you store objects in the Svelte store and call the `set`, all the listeners will be notified about the update, even when the value hasn't changed ([source](https://arc.net/l/quote/sveijxhx)):
> The store then checks whether the new value differs from the previous before notifying all of its subscribers. However, when the store contains an array or object, it _skips the check_, as seen from [the comparing algorithm](https://github.com/sveltejs/svelte/blob/670f4580568fe8ea31097981ba2d59c33daf0725/src/runtime/internal/utils.ts)

## Issue
If you subsequently call the `selection.clear()`, it will notify the `selection.subscribe` callback each time. That causes React components that read from the `useSelection` hook re-render each time too.

## Changes Made
Added a simple comparison before clearing the selection state. So calling the `selection.clear()` will notify the listeners only once.

### DIFF
--- a/packages/annotorious-core/src/state/Selection.ts
+++ b/packages/annotorious-core/src/state/Selection.ts
@@ -1,4 +1,6 @@
 import { writable } from 'svelte/store';
+import { dequal } from 'dequal/lite';
+
 import type { Annotation } from '../model';
 import type { Store } from './Store';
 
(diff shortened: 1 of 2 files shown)


rsimon:
Looks good, thanks!

PR #443, opened 2024-09-12
Add point tool

RomainDECOSTER:
Hello,

I hope you are doing well.

I was using version 2.7 and in this version there was the point tool that I used a lot. So I thought I would add it.

Best Regards,

### DIFF
--- a/packages/annotorious/src/annotation/editors/point/PointEditor.svelte
+++ b/packages/annotorious/src/annotation/editors/point/PointEditor.svelte
@@ -1,9 +1,5 @@
 <script lang="ts">
-  import {
-    boundsFromPoint,
-    boundsFromPoints,
-    POINT_RADIUS,
-  } from '../../../model';
+  import { boundsFromPoint, POINT_RADIUS } from '../../../model';
   import type { Point, PointGeometry, Shape } from '../../../model';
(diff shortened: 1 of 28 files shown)


packages/annotorious/src/annotation/editors/point/PointEditor.svelte
```diff
@@ -0,0 +1,55 @@
...
+      ...point,
+      geometry: { ...pointTranslated, bounds },
+    };
+  };
+</script>
+
+<Editor {shape} {transform} {editor} on:change on:grab on:release let:grab>
+  <circle
+    class="a9s-outer"
+    style="fill:yellow;stroke:darkgoldenrod;"
```
RomainDECOSTER: I don't know how to inform the user that the point is selected.

RomainDECOSTER: I found:
CleanShot 2024-09-12 at 16 41 11@2x

rsimon:
Thanks for this! I'll try to revise some things a little (might happen over the weekend), but that's really helpful! Also, I agree it makes sense to have a point tool as part of the base package. (In v2.7, the point tool was in the extensions packages.)

PR #459, opened 2024-10-25
Compatible with dynamic importing in Next.js

Songhao-LI:


rsimon:
Thanks for looking into this! It seems odd that this works, because technically that config isn't actually correct: "require" should always point to a UMD build, not an ES build (which is the case for `./dist/annotorious-react.es.js`). I'm worried that this might break things in other setups, or with other bundlers...

I googled a bit, and it looks like the following could be a cleaner solution:

```
"exports": {
  ".": {
    "types": "./dist/index.d.ts",
    "import": "./dist/annotorious-react.es.js",
    "default": "./dist/annotorious-react.es.js"
  },
  "./annotorious-react.css": "./dist/annotorious-react.css"
},
```

I.e. set the ES build as a `default`, rather than the target for `require`. Can you confirm that this works on  your end?

Songhao-LI:
cool! it works now!
Thank you for your time, and I've updated it in pr as well

Songhao-LI:
> Thanks for looking into this! It seems odd that this works, because technically that config isn't actually correct: "require" should always point to a UMD build, not an ES build (which is the case for `./dist/annotorious-react.es.js`). I'm worried that this might break things in other setups, or with other bundlers...
> 
> I googled a bit, and it looks like the following could be a cleaner solution:
> 
> ```
> "exports": {
>   ".": {
>     "types": "./dist/index.d.ts",
>     "import": "./dist/annotorious-react.es.js",
>     "default": "./dist/annotorious-react.es.js"
>   },
>   "./annotorious-react.css": "./dist/annotorious-react.css"
> },
> ```
> 
> I.e. set the ES build as a `default`, rather than the target for `require`. Can you confirm that this works on your end?

cool! it works now!
Thank you for your time, and I've updated it in pr as well.

And it seems like there are still some compatibilities when using Next.js with dynamic importing
I have created a demo here: https://codesandbox.io/p/github/Songhao-LI/AnnotoriousExample/main
you can uncomment the OpenSeadragonAnnotationPopup component to view the error when you finish creating an annotation.

rsimon:
Thanks for this!

Hm, no idea why this error would happen. The problem is in this line:
https://github.com/annotorious/annotorious/blob/main/packages/annotorious-react/src/openseadragon/OpenSeadragonAnnotationPopup.tsx#L98

The message seems to indicate that the `viewer` is null, which doesn't make sense given that the popup is opened from an annotation (on the viewer). Not sure if NextJS SSR messes with the React context somehow. (The `viewer` is consumed by the popup through the Annotorious context.)

Songhao-LI:
> Thanks for this!
> 
> Hm, no idea why this error would happen. The problem is in this line: https://github.com/annotorious/annotorious/blob/main/packages/annotorious-react/src/openseadragon/OpenSeadragonAnnotationPopup.tsx#L98
> 
> The message seems to indicate that the `viewer` is null, which doesn't make sense given that the popup is opened from an annotation (on the viewer). Not sure if NextJS SSR messes with the React context somehow. (The `viewer` is consumed by the popup through the Annotorious context.)

Thanks!!

Is there any alternative way to integrate a custom popup in Annotorious 3.x, similar to how it worked in Annotorious 2.x? Specifically, I guess adding the popup wiget by assigning properties to the Annotorious instance itself might solve this problem since we can pass in the reference of viewer.

rsimon:
I'm afraid that wouldn't change anything. The root problem isn't the extension mechanism. It's the fact that Next seems to swallow the reference to the OpenSeadragon viewer. I've seen similar behavior in cases where two copies of OpenSeadragon were in the same page. Are you sure there isn't a problem with the way Next bundles and hydrates the code? In Astro, there's a directive that forces code to be left alone on the server, and only get touched in the browser. Perhaps there are similar tuning levers that could be used to tweak Next's behavior?

Songhao-LI:
> I'm afraid that wouldn't change anything. The root problem isn't the extension mechanism. It's the fact that Next seems to swallow the reference to the OpenSeadragon viewer. I've seen similar behavior in cases where two copies of OpenSeadragon were in the same page. Are you sure there isn't a problem with the way Next bundles and hydrates the code? In Astro, there's a directive that forces code to be left alone on the server, and only get touched in the browser. Perhaps there are similar tuning levers that could be used to tweak Next's behavior?


Thanks for the information!! I will have a try

rsimon:
Ouch - I just looked at your code again. Your problem isn't related to Next, React contexts, hydration, etc. at all. I just realized you have placed the popup component outside of the annotator component. 

You just have to put the OpenSeadragonAnnotationPopup as a child of the OpenSeadragonAnnotator :-)

Songhao-LI:
> Ouch - I just looked at your code again. Your problem isn't related to Next, React contexts, hydration, etc. at all. I just realized you have placed the popup component outside of the annotator component.
> 
> You just have to put the OpenSeadragonAnnotationPopup as a child of the OpenSeadragonAnnotator :-)

I see!! Thanks!!
I just follow the example here, which seems not correct. Maybe you can revise later 👍 
image

rsimon:
Ah - indeed!! Fixed!

PR #467, opened 2024-11-04
Added `isIdling` to the `Selection` state

oleksandr-danylchenko:
## Issue
See - https://github.com/recogito/text-annotator-js/issues/172. When a user rapidly selects the text down - their touch can go over the floating popup element which causes the highlight to flicker.
To prevent that, we decided to make the custom popup work similarly to the native context menu. It will appear only once the selection idles. 

## Changes Made
To represent that state I added the `isIdling` property to the `Selection` state. The value will automatically become accessible via the `useSelection` hook and will be reactive due to the Svelte `writable` properties.

#### Caveat
The `isIdling` value will only be utilized by the `r6o/text-annotator-js` for now. It already has a use-case for there and a relatively clear way to manage it.

### DIFF
--- a/packages/annotorious-core/src/state/Selection.ts
+++ b/packages/annotorious-core/src/state/Selection.ts
@@ -9,6 +9,8 @@ export interface Selection {
 
   event?: PointerEvent | KeyboardEvent;
 
+  isIdling?: boolean;
+
 }
 
 export type SelectionState<I extends Annotation, E extends unknown> = ReturnType<typeof createSelectionState<I, E>>;
@@ -49,8 +51,7 @@ export const createSelectionState = <I extends Annotation, E extends unknown>(
   const isEmpty = () => currentSelection.selected?.length === 0;
 
   const isSelected = (annotationOrId: I | string) => {
-    if (isEmpty())
-      return false;
+    if (isEmpty()) return false;
 
     const id = typeof annotationOrId === 'string' ? annotationOrId : annotationOrId.id;
     return currentSelection.selected.some(i => i.id === id);
@@ -87,7 +88,8 @@ export const createSelectionState = <I extends Annotation, E extends unknown>(
 
     set({
       selected: annotations.map(annotation => {
-        // If editable is not set, use default behavior
+
+        // If editable isn't set, use the default behavior
         const isEditable = editable === undefined
           ? onUserSelect(annotation, currentUserSelectAction, adapter) === UserSelectAction.EDIT
           : editable;
@@ -112,6 +114,12 @@ export const createSelectionState = <I extends Annotation, E extends unknown>(
       set({ selected: selected.filter(({ id }) => !ids.includes(id)) });
   }
 
+  const setIdling = (isIdling: boolean) => {
+    if (currentSelection.isIdling !== isIdling) {
+      set({ ...currentSelection, isIdling });
+    }
+  }
+
   const setUserSelectAction = (action: UserSelectActionExpression<E> | undefined) =>
     currentUserSelectAction = action;
 
@@ -130,13 +138,17 @@ export const createSelectionState = <I extends Annotation, E extends unknown>(
     get userSelectAction() {
       return currentUserSelectAction;
     },
-    clear,
+    get isIdling() {
+      return currentSelection ? currentSelection.isIdling : null;
+    },
     isEmpty,
     isSelected,
     setSelected,
+    setIdling,
     setUserSelectAction,
+    userSelect,
     subscribe,
-    userSelect
+    clear
   };
 
 }
--- a/packages/annotorious-core/src/state/Selection.ts
+++ b/packages/annotorious-core/src/state/Selection.ts
@@ -9,8 +9,6 @@ export interface Selection {
 
   event?: PointerEvent | KeyboardEvent;
 
-  [key: string]: any; // Allow for additional properties to be added by plugins
-
 }
 
 export type SelectionState<I extends Annotation, E extends unknown> = ReturnType<typeof createSelectionState<I, E>>;
@@ -21,7 +19,7 @@ export enum UserSelectAction {
 
   SELECT = 'SELECT',  // Just select, but don't make editable
 
-  NONE = 'NONE' // Click won't select - annotation is completely inert
+  NONE = 'NONE' // Click won't select - the annotation is completely inert
 
 }
 


oleksandr-danylchenko:
The only byproduct of such a change is that the `r.on('selectionChanged', ...` handler gets called twice: 
image
I'll how/whether it can be conveniently filtered out in the `Lifecycle` handler.

rsimon:
Wait wait wait - not so fast ;-) You are cutting through literally ALL the layers not just of the text annotator, but Annotorious, too. But what we want is simple: we just want to react to changes in the annotation target, in the TextAnnotationPopup. If the target changes, we hide the popup, and show it again after short time period.

It's really a single concern, and we can handle it centrally in the place where we need it - the TextAnnotationPopup component. No need to touch the selection handler, let alone the Annotorious core.

oleksandr-danylchenko:
Yeah, you're right...

Reacting to the changes and showing the popup over a short delay is definitely a pragmatic and simple enough option. It's not as "robust" or "controlled" as broadcasting it via the centralized `Selection` state. But I support the latter one is a bit of an overengineering edge. I am not sure what caused me not to consider a simpler solution first 😅

oleksandr-danylchenko:
Or... Even a more versatile strategy might be extracting the "idle" state listening to the hook! Because I need such a state in both the `TextAnnotatorPopup`,  and the [`useAnnouncePopupNavigation`](https://github.com/recogito/text-annotator-js/pull/141/files#diff-86cf4735dfec418174559a308bca544a80c6178258372614d602a699166b0602R56-R73) ([discussion source](https://github.com/recogito/text-annotator-js/issues/172#issuecomment-2455076827)).

rsimon:
Nice - sounds like a good solution. Keeps the logic in a single place, yet makes it reusable!

oleksandr-danylchenko:
After trying to fit a generic "listen to the store updates" approach into a hook I faced a few challenges 🤔 
A user can halt their selection for a second or two, w/o lifting the pointer. However, that will make the `store` think that the selection is "idling" and we should be good to display the popup back. So effectively it requires binding the "idling" state to the `SelectionHandler` events 🤔 

---

Here's the `useSelectionIdling` hook example:
```tsx
export const useSelectionIdling = (options: { delay: number } = { delay: 200 }) => {
  const store = useAnnotationStore();

  const [isIdling, setIdling] = useState(true);

  useEffect(() => {
    if (!store) return;

    let idleTimeout: ReturnType<typeof setTimeout>;

    const scheduleIdleSetting = () => {
      clearTimeout(idleTimeout);
      idleTimeout = setTimeout(() => setIdling(true), options?.delay);
    };

    const handleStoreUpdate = () => {
      setIdling(false);
      scheduleIdleSetting();
    };
    store.observe(handleStoreUpdate, { ignore: Ignore.BODY_ONLY, origin: Origin.LOCAL });

    return () => {
      clearTimeout(idleTimeout);
      store.unobserve(handleStoreUpdate);
    };
  }, [store]);

  return isIdling;
};
```
See how the `isIdling` state jumps back and forth when I halt the selection for a bit:

https://github.com/user-attachments/assets/30879fb3-68e9-4622-8901-a625175985c6

rsimon:
Not sure I understand. You'd hide the popup on annotation change events, right? And maintain a timeout for re-displaying the popup after a period of time where no change events fired?

oleksandr-danylchenko:
Here's the `useSelectionIdling` hook example:
```tsx
export const useSelectionIdling = (options: { delay: number } = { delay: 200 }) => {
  const store = useAnnotationStore();

  const [isIdling, setIdling] = useState(true);

  useEffect(() => {
    if (!store) return;

    let idleTimeout: ReturnType<typeof setTimeout>;

    const scheduleIdleSetting = () => {
      clearTimeout(idleTimeout);
      idleTimeout = setTimeout(() => setIdling(true), options?.delay);
    };

    const handleStoreUpdate = () => {
      setIdling(false);
      scheduleIdleSetting();
    };
    store.observe(handleStoreUpdate, { ignore: Ignore.BODY_ONLY, origin: Origin.LOCAL });

    return () => {
      clearTimeout(idleTimeout);
      store.unobserve(handleStoreUpdate);
    };
  }, [store]);

  return isIdling;
};
```
See how the `isIdling` state jumps back and forth when I halt a selection for a bit:

https://github.com/user-attachments/assets/30879fb3-68e9-4622-8901-a625175985c6

oleksandr-danylchenko:
> You'd hide the popup on annotation change events, right? And maintain a timeout for re-displaying the popup after a period of time where no change events fired?

Yes, that's correct. But the "pause" in the change event emission can still happen when the selection is "technically active"; when a user hasn't released a finger from the handlebar or didn't release a mouse.

rsimon:
Yes, but as long as the user doesn't move, it's ok to display the popup - even if the finger is down?

oleksandr-danylchenko:
> it's ok to display the popup - even if the finger is down?

Technically, it fixes the original bug we had. A user won't go over the popup. But it makes the popup act not the native context menu. The latter appears when the user actually "concludes" the selection by releasing their finger. So with binding to the `SelectionHandler` events I was striving toward such a behavior too.
Plus, that can make the popup appear a bit annoying when it constantly hides/appears during a single "selection cycle" 🤔 

For example (using a9s example just for the convenience): 

https://github.com/user-attachments/assets/d6b68f6b-6cc7-4f48-9471-19be7775150e

rsimon:
Personally, I wouldn't see it as a problem. But the hook could still track pointer activity in addition. (Keep in mind that selection handling works different between image and text. There's no direct selection handler equivalent in Annotorious.)

oleksandr-danylchenko:
> But the hook could still track pointer activity in addition

Yeah, that sounds like an option to me. We can add a generic hook in A9S that will track just the store's idling first. And then in the R6O, add a hook that will not only listen to the store but to the text-specific events like `selectionchange`, `pointerup`, `contextmenu`, etc. In that way, it promises to be a reliable way to track the "concluded" selection and when the popup should be displayed.

rsimon:
I honestly wouldn't split it up between more places than absolutely necessary, prematurely. I agree that a popup that goes away if the user edits the shape and re-appears after pointerup would be useful for image as well. But to keep it simple initially, I'd start by simply building this as a single hook, in the text annotator project (given that it will have to handle text-specific things like `selectionchange` and `contextmenu` as you say).

oleksandr-danylchenko:
Ugh... That turned out to be quite a rabbit hole...

Firstly, we cannot simply react to the `store` or `selectionchange` events. That's because they are also triggered on a desktop that doesn't have a notion of the "context menu" and there can be no overlap with the popup. So simply adding the timeout-based hiding will make the popup flicker back and forth on keyboard selection.

Ideally, we need to track the native context menu appearance, but it's another challenge...
The tracking should start only when there is _an actual touch_. But then, if we're outside the `SelectionHandler` it requires to re-implement a state-machine that will follow such transitions: `pointerdown - > selectionchanged -> pointerup`. However, it won't work on Android that doesn't emit `pointerup`, we'll need an additional listener for the `contextmenu`. All that makes the native context tracking super wanky.

But I'm still thinking about the workaround 🙏🏻

oleksandr-danylchenko:
That's crazy... I found that when you make the selection for the 1st time, the `contextmenu`'s `event.pointerType` is the `touch`, as expected. But when you proceed with the selection and release the handlebar again - it's the `mouse`! 🤯 

image
image

oleksandr-danylchenko:
Okay... I'm giving up on it... 😔
The `selectionchange` + `contextmenu` event combination is crazy. The selection can correct itself when finished of a whitspace. So even after the context menu showed up, that will emit the trailing `selectionchange`. And as we want to "hide" the context menu on selection change - that will lead to an incorrect state. 
Furthermore, I couldn't find a predictable timing connection between the `contextmenu` and `selectionchange`. When it's a _one-word selection_, the difference between them is _from 10ms. to 41ms._ But the correction over the whitespace varies from `360ms.` to `470ms.`
It all could be much easier if could track `pointerdown`/`pointerup` events when user touches/drags the handlebars - but _those events are missing_.

---

Therefore, I think that we should go with a not that ideal but yet pragmatical solution instead. We should listen for the `store` changes for the selected annotation's `target` and display the popup only when it's idling for ~1s.

PR #471, opened 2024-11-06
Fixed local test image reference

oleksandr-danylchenko:
## Issue
Unfortunately, using the previous `img` declaration, it was producing a broke image: 
image

## Change Made
Added proper local asset referencing via the `public` folder: 
image


packages/annotorious-react/package.json
```diff
@@ -8,7 +8,8 @@
   "funding": "https://steadyhq.com/rainer-simon",
   "type": "module",
   "scripts": {
-    "start": "vite",
+    "start-image": "cd test/image && vite --host",
+    "start-openseadragon": "cd test/openseadragon && vite --host",
```
oleksandr-danylchenko, line 12: Just a single `vite` picks the last `index.html` under the `test` folder hierarchy. Effectively running only the `openseadragon` test

rsimon:
Cannot reproduce... in the `react` package, running `npm start` opens `http://localhost:5173/test/openseadragon/index.html` for me, which works correctly. Changing to `http://localhost:5173/test/image/index.html` brings up the image test page for me?

oleksandr-danylchenko:
Interesting... It really works when I start the app from the root of the package. But it stops working when I run `vite` from within the `tests/image` folder:

https://github.com/user-attachments/assets/6122d8c1-7fcd-4d54-8f8d-1aaa3df8bd30

https://github.com/user-attachments/assets/4bcce711-c8b2-4fd1-be79-4854a00a3bc0

oleksandr-danylchenko:
IMHO, having separate scripts instead of manually navigating the tests via URL feels more obvious for the devs 🤔 
So this PR address the image issue present for the "separate scripts"

PR #482, opened 2024-11-29
Svelte 5 migration

bmschmidt:
Initial stab at migrating core components and types to svelte 5. Not remotely ready for merge, but at a point where I wouldn't want to put in any more without direction from @rsimon if you think it's worth pursuing.

Notes:

1. I'm not sure if this works or not.
2. I stayed mostly in the code checking types rather than addressing functionality. Most likely I turned something off in addition to the shortcomings loaded below.
3. I'm able to load `http://localhost:5173/test/index.html` without console errors and with the shapes showing up and `npm test` succeeds.
4. I don't think I've tested the editor components.
5. Remaining errors on running `npx svelte-check` are primarily around the OSD package, which I didn't attempt to fully migrate, and some type errors on passing the editor/ that I think may be harmless based on [this thread](https://github.com/storybookjs/storybook/issues/29308).
6. The diff is larger than it needs to be because prettier did some changes when I saved, sorry. If desired I could put a separate PR first that adds a [`pretty-quick`](https://www.npmjs.com/package/pretty-quick) hook closer to the project style.
7. I commented out the bits relating to registering new plugins because it was hard and I don't completely understand the use case. I have some fuzzy ideas about how to expose interfaces there but wasn't sure if it was in-bounds to tweak the external plugin API at all.
8. Left a number of TODOs scattered throughout.


packages/annotorious/src/index.ts
```diff
@@ -3,7 +3,7 @@ export * from './model';
 export * from './state';
 export * from './themes';
 export * from './annotation/tools';
-export * from './Annotorious';
+export * from './Annotorious.svelte';
```
bmschmidt, line 6: This turned out not to be necessary, but at some point would be useful to do this because the stores could be replaced with reactive state.

packages/annotorious/src/annotation/tools/polygon/RubberbandPolygon.svelte
```diff
@@ -129,63 +147,62 @@
...
-    addEventListener('dblclick', onDblClick, true);
-  });
+  };
+  console.error('HERE');
+  // onMount(() => {
+  //   addEventListener('pointerdown', onPointerDown, true);
+  //   addEventListener('pointermove', onPointerMove);
+  //   addEventListener('pointerup', onPointerUp, true);
+  //   addEventListener('dblclick', onDblClick, true);
+  // });
```
bmschmidt, line 186: This is all TODO here, forgot about this.

rsimon:
Wow, thanks, this is fantastic! Although it does somewhat confirm my initial fear that the effort required for migration is heavy... I'll try to look over it in more detail over the weekend!

packages/annotorious/src/annotation/tools/drawingToolsRegistry.ts
```diff
@@ -1,9 +1,9 @@
-import type { SvelteComponent } from 'svelte';
-import { RubberbandRectangle } from './rectangle';
-import { RubberbandPolygon } from './polygon';
+import RubberbandRectangle from './rectangle/RubberbandRectangle.svelte';
+import RubberbandPolygon from './polygon/RubberbandPolygon.svelte';
 import type { DrawingMode } from '../../AnnotoriousOpts';
+import type { ShapeName } from 'src/model';
 
-export type DrawingTool = 'rectangle' | 'polygon' | string;
+export type DrawingTool = 'RECTANGLE' | 'POLYGON';
```
rsimon, line 6: This is something we'd need to keep the same, in order not to break the consuming apps. The `| string` isn't ideal either I admit. But the basic idea is that plugins are going to register new tools (with new names), so I don't see a way to keep this strongly typed without breaking extensibility.

packages/annotorious/src/annotation/tools/drawingToolsRegistry.ts
```diff
@@ -13,15 +13,21 @@ export type DrawingToolOpts = {
...
-  
-export const registerTool = (name: string, tool: typeof SvelteComponent, opts?: DrawingToolOpts) =>
-  REGISTERED.set(name, { tool, opts });
+export const getTool = (name: ShapeName) => {
+  if (name === "ELLIPSE") {
+    throw new Error(`Not supported right now. Seems like this is part
+      of an extension that needs to be baked into the main type system?`)
+  }
+  // TODO: I don't understand 'opts'
+  return {tool: drawingTool[name], opts: undefined}
```
rsimon, line 29: The `opts` are additional config that a plugin can register along with the tool.

packages/annotorious/src/annotation/tools/drawingToolsRegistry.ts
bmschmidt: Hmm I think in that case it should probably just be string. I changed this because there are some benefits to having the tool be the same as the shape name, but on reflection that probably is just wrong -- no reason to assume that there is only one way to edit polygon shapes.

rsimon: > no reason to assume that there is only one way to edit polygon shapes.

My thoughts exactly when designing this! You can have different tools that produce the same shape type. (Annotorious v2 already did.) That's why there's also a separation between the "runtime rendering" of the static shape as such, and the editable representation via tools/editors. (This could be designed in a more extensible way I guess. But then... it works & is reasonably simple.)

I agree the type could just be a string. Originally, I added `rectangle` and `polygon` to get a bit of IDE auto-completion for the built-in tools/editors, although it doesn't actually work that way. (For odd reasons, it would have to be `| string & {};`. So that's probably the way to go, but I never actually added this...)

rsimon:
I decided that I'll drop Svelte from Annotorious in the mid-/long run, and instead go with [SolidJS](https://www.solidjs.com/) for the next major release. Thanks for the help @bmschmidt! But ultimately I guess my excitement with Svelte has worn off ;-)

PR #487, opened 2024-12-06
Default cusor styling if user selection is set to none

ysabri:
Hi, 

I ran into this while wanting to display the annotations to users who aren't allowed to edit them. I hope I did it right, lmk if I need to change anything.


rsimon:
Superseded by #618

Issue #493, opened 2024-12-13
The `userSelectAction` change doesn't lead to selection state re-evaluation


oleksandr-danylchenko:
## **Issue behavior**
The selection state evaluation is based on the `currentUserSelectAction` to assign whether an annotation is editable. https://github.com/annotorious/annotorious/blob/94624c0dea045db435d30ba054a29e5c97f83d22/packages/annotorious-core/src/state/Selection.ts#L79-L89
https://github.com/annotorious/annotorious/blob/94624c0dea045db435d30ba054a29e5c97f83d22/packages/annotorious-core/src/state/Selection.ts#L103-L106

However, when a consumer changes the `userSelectionAction`, the selection state won't be updated respectively: https://github.com/annotorious/annotorious/blob/94624c0dea045db435d30ba054a29e5c97f83d22/packages/annotorious-core/src/state/Selection.ts#L128-L129

**That way, a user can create an "impossible" state where the `userSelectAction` is `SELECT` or `NONE`, but both the `selected` is present and the `editable` is `true`.**

PR #494, opened 2024-12-13
#493 Added selection state re-evaluation on `userSelectAction` change

oleksandr-danylchenko:
## Issue
See - https://github.com/annotorious/annotorious/issues/493

>  A user can create an "impossible" state where the `userSelectAction` is `SELECT` or `NONE`, but both the `selected` is present and the `editable` is `true`.** 

## Changes Made
Added selection state re-evaluation call after updating the  `currentUserSelectAction`. That way, the `editing` property on the selection will stay up-to-date. Also, the consumers subscribed to the selection state will receive the updated state properly.


rsimon:
Good catch - thanks!

PR #495, opened 2024-12-16
SVG path parser

flamsens:
This PR adds a SVG path parser to the W3C adapter. The parser supports single line paths only and outputs a polygon shape.


rsimon:
That's fantastic, thanks so much! I'll test this when I have a bit more time - but at first glance this looks great.

Just a (side) question: did you have to add the Svelte config to make the build work in your case?

flamsens:
When I clone the annotorious repository, I can't run the tests on my machine (node 22.11.0, npm 10.9.0). Adding the svelte plugin fixed the problem in my case. I have no experience with Svelte, so it might be some other problem ...

image

PR #500, opened 2024-12-19
Fix: Annotation style rgba alpha values are ignored by Pixi renderer

flamsens:
If you use a rgba fill/stroke color in the annotation style, the alpha values are ignored. The Pixi stageRenderer only uses fillOpacity/strokeOpacity or the default values of the renderer.

```javascript
{
  fill: hovered ? 'rgba(0,255,0,0.1)' : 'rgba(255,0,0,0.2)', # alpha value are always ignored by pixi renderer
  stroke: '#ff0000', 
  // fillOpacity: 0.1,
  strokeOpacity: 0.9,
  strokeWidth: 1
}
```


rsimon:
That's a nice enhancement - thanks!

PR #511, opened 2025-01-22
Fix/osd viewer reload

### LINKED ISSUE #508: Working with OSD with TileSource from GeoTIFFTileSource
**Which version of Annotorious are you using?**
Please provide:
"@annotorious/openseadragon": "^3.0.19",
"openseadragon": "^5.0.1",

**Issue behavior**
I am unable to create annotations when using TileSource from [GeoTIFFTileSource](https://github.com/pearcetm/GeoTIFFTileSource). As soon as I annotate, the whole OSD viewer seems to refresh. Has anyone faced this issue? I can attach an example Geotiff file if that helps.

**Screenshots**

devbharat:
Fixes #508 

Tested OK with images and geotifFTileSource.

Note: Might break the viewer for people who who properly memoize their OpenSeadragon options object.

### DIFF
--- a/packages/annotorious-react/src/openseadragon/OpenSeadragonViewer.tsx
+++ b/packages/annotorious-react/src/openseadragon/OpenSeadragonViewer.tsx
@@ -28,13 +28,14 @@ export const OpenSeadragonViewer = forwardRef<OpenSeadragon.Viewer, OpenSeadrago
         setViewer(v);
 
       return () => {
+        console.log('Destroying viewer', JSON.stringify(options));
         if (setViewer)
           setViewer(undefined);
 
         v.destroy();
       }
     }
-  }, [JSON.stringify(options)]);
+  }, [options]);
 
   useImperativeHandle(ref, () => viewer);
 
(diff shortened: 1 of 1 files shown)


Clasen00:
Hello! Thank you for your work!
I just noticed that this PR breaking my project. I got an error: `Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.` 
It happened because my options are object therefor it cannot be used as a dependency in useEffect
image
this is the place where console is point to:
image
@rsimon @devbharat

devbharat:
The viewer effect now keys off the options object by reference. If you pass a new object each render you’ll trigger a re‑init loop. Please wrap your options in useMemo (and avoid in‑place mutation):

const options = useMemo(() => ({ tileSources, preserveViewport: true, /*...*/ }), [tileSources]);
<OpenSeadragonViewer options={options} />

Clasen00:
@devbharat Thank you! thats work for me, but its not so obviously now :)

rsimon:
Ah - yes, sorry. This did actually slip into the 3.7 release more by accident than by intention. 

But, yes, as @devbharat says (thanks!): the proper way to handle this is to memo-ize the options object, and there are some cases where keying off the object by reference might be preferable. I'll update the docs accordingly!

Clasen00:
@rsimon okay, thats not so bad:) Still a great work, thank you!

Issue #496, opened 2024-12-16
The `useAnnotationSelectAction` doesn't respect the annotator's `adapter`


oleksandr-danylchenko:
## Issue behavior
See - https://github.com/annotorious/annotorious/commit/79c7b17f2002ff2dfcc93880c9048c8a73ad01e2#r150398933

In the `useAnnotationSelectAction` hook, when the `onUserSelect` is called - it doesn't receive the `adapter` associated with the annotator instance.
However, the `UserSelectActionExpression<E>` type dictates that the selection action callback will run on the external type annotation instance.
https://github.com/annotorious/annotorious/blob/94624c0dea045db435d30ba054a29e5c97f83d22/packages/annotorious-react/src/Annotorious.tsx#L164-L167

That makes the user create the `userSelectAction` callback that will throw the **_runtime exception_** while trying to run the `useAnnotationSelectAction`.

## Screenshots
![image](https://github.com/user-attachments/assets/2af167ce-8f0e-4665-a2da-dc3372d32237)
However, in runtime, the `userSelectAction` will receive an internal type instance:
![image](https://github.com/user-attachments/assets/be4afb5f-366b-42d9-8771-586d733c7a3f)

PR #517, opened 2025-02-21
#496 Added missing `adapter` to the `useAnnotationSelectAction`

oleksandr-danylchenko:
## Issue
See - https://github.com/annotorious/annotorious/issues/496

## Changes Made
Added `adapter` argument to the `useAnnotationSelectAction` to prevent type mismatch in the runtime. It follows the signature of the underlying `onUserSelect`:
https://github.com/annotorious/annotorious/blob/fb4e7848870ea627998f910284eeb3cce2f214fe/packages/annotorious-core/src/state/Selection.ts#L159-L166


rsimon:
Thanks for this! I was sort of hoping that the adapter could go in automatically (from the main `props` on the annotator component). But that would probably be tricky for TS. Perhaps this is the more explicit solution, anyway?

oleksandr-danylchenko:
> could go in automatically

Yeah, it definitely would be more convenient 🤔
The only "significant" change required for it would be exposing the `adapter` property on the `annotator` instance. 

Moreover, that might be a _rather needed option_ for React consumers, especially the `react-text-annotator`. As there, the `adapter` might be declared as the expression ([source](https://github.com/recogito/text-annotator-js/blob/5b9abad31724db14d473f7d363b295a967cd93b7/packages/text-annotator-react/src/TextAnnotator.tsx#L13)): 
```ts
adapter?: FormatAdapter<I, E> | ((container: HTMLElement) => FormatAdapter<I, E>) | null;
```
Therefore, if the app uses the expression, _it won't be able to pass it to the `useAnnotationSelectAction`!_

I'll give this PR another look and will try to make the `adapter` injection automatic 🙌🏻

oleksandr-danylchenko:
With the new `evalSelectAction` (https://github.com/annotorious/annotorious/commit/f8be650c328008c45764dacdc546362f225b662f) it would be easier to implement the hook w/o the need to always pass the `adapter`.

packages/annotorious-react/src/Annotorious.tsx
```diff
@@ -86,56 +77,56 @@ export const Annotorious = forwardRef<Annotator, { children: ReactNode }>((props
...
 
 export const useAnnotator = <T extends unknown = Annotator<any, unknown>>() => {
   const { anno } = useContext(AnnotoriousContext);
   return anno as T;
-}
+};
 
 export const useAnnotationStore = <T extends unknown = Store<Annotation>>() => {
-  const { anno } = useContext(AnnotoriousContext);
+  const anno = useAnnotator();
```
oleksandr-danylchenko, line 111: The `anno` returned directly from the `AnnotoriousContext` is untyped. Instead, we can use the typed wrapper - `useAnnotator`.

packages/annotorious-react/src/Annotorious.tsx
```diff
@@ -159,20 +150,23 @@ export const useAnnotation = <T extends Annotation = ImageAnnotation>(
...
-export const useAnnotationSelectAction = <I extends Annotation = ImageAnnotation, E extends unknown = ImageAnnotation>(id: string, action: UserSelectActionExpression<E>) => {
+export const useAnnotationSelectAction = <I extends Annotation = ImageAnnotation>(
+  id: string
+) => {
+  const anno = useAnnotator();
   const annotation = useAnnotation<I>(id);
-  return annotation ? onUserSelect<I, E>(annotation, action) : undefined;
-}
+  return anno && annotation ? anno.state.selection.evalSelectAction(annotation) : undefined;
+};
```
oleksandr-danylchenko, line 161: With the `evalSelectAction`, we can simplify the `useAnnotationSelectAction` signature to just the single `id` argument. The rest will be prepopulated from the `evalSelectAction` closure.

Issue #518, opened 2025-02-21
Why "Updates are only applied immediately if they involve body changes"?


oleksandr-danylchenko:
@rsimon, could you please clarify the reason behind limiting the immediate `updateAnnotation` emission if the annotation doesn't include the bodies?
https://github.com/annotorious/annotorious/blob/fb4e7848870ea627998f910284eeb3cce2f214fe/packages/annotorious-core/src/lifecycle/Lifecycle.ts#L159-L164

---

In my app, I have a use case where a user can update the annotation-level property, click save, and expect that the `updateAnnotation` will get called to persist it on the server:
```ts
// note's form
...
if (isSharedChanged || isValueChanged) {
	store!.updateAnnotation(newAnnotation);
}
```
```ts
// API updates listener
...
const handleCreateOrUpdateAnnotation = (annotation: W3CWebtextsAnnotation) => {
	if (canCreateOrUpdateApiAnnotation(annotation)) {
		api.createOrUpdate?.(annotation);
	}
};

r.on('createAnnotation', handleCreateOrUpdateAnnotation);
r.on('updateAnnotation', handleCreateOrUpdateAnnotation);
```
Unfortunately, due to the gating above, only the body updates are being propagated immediately upon calling the `updateAnnotation`.

https://github.com/user-attachments/assets/94e4bee7-5d21-4e74-8e26-d014b69c4353

That requires a user to change the selected annotation to persist the change:
https://github.com/annotorious/annotorious/blob/fb4e7848870ea627998f910284eeb3cce2f214fe/packages/annotorious-core/src/lifecycle/Lifecycle.ts#L102-L114

rsimon:
It's to prevent excessive update events when users drag an image annotation, thus producing lots of target changes. (The lifecycle events are specifically for basic CRUD operations, rather than realtime updates.)

oleksandr-danylchenko:
> It's to prevent excessive update events when users drag an image annotation, thus producing lots of target changes

Thanks! I didn't think about that detail. As in the text annotator the `target` remains almost static upon the initial selection.

oleksandr-danylchenko:
What do you think about changing the API layer listeners from the `r.on('...')` to the `store.observe`? Wouldn't it be too hacky tapping into an "internal" method of the `Annotator` (as the annotations reported via the `.observe` don't go through the `adapter`)?

> (pseudo-code):
> ```ts
> store.observe(event => {
>   // Local CREATE and DELETE events are applied immediately
[...]

PR #519, opened 2025-02-24
#518 Added `immediateUpdatesEmit` option to the Lifecycle observer

oleksandr-danylchenko:
## Issue
See - https://github.com/annotorious/annotorious/issues/518#issuecomment-2678403410

## Changes Made
Added the `immediateUpdatesEmit` option to the lifecycle observer. It makes the store observer bypass the body update presence check and emit the `updateAnnotation` even immediately.

## Relations
This PR is required for the:
- https://github.com/recogito/text-annotator-js/pull/193


oleksandr-danylchenko:
See - https://github.com/annotorious/annotorious/issues/518#issuecomment-2678844514

PR #521, opened 2025-03-03
Add setDrawingEnabled to Annotator type definition

willemarcel:
The Annotator type definition is missing the setDrawingEnabled property

### DIFF
--- a/packages/annotorious-core/src/model/Annotator.ts
+++ b/packages/annotorious-core/src/model/Annotator.ts
@@ -54,6 +54,8 @@ export interface Annotator<I extends Annotation = Annotation, E extends unknown 
 
   setAnnotations(annotations: Partial<E>[], replace?: boolean): void;
 
+  setDrawingEnabled(enabled: boolean): void;
+
   setFilter(filter: Filter<I> | undefined): void;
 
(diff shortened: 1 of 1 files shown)


rsimon:
Hi, the fact that the annotator doesn't have this method is intentional. It's a base class that's also re-used with other media annotators–such as a text annotator that doesn't have drawing capability. The method is defined at the level of the [ImageAnnotator](https://github.com/annotorious/annotorious/blob/main/packages/annotorious/src/Annotorious.ts).

willemarcel:
Ah! I just discovered that I can pass the drawingEnabled property to the `ImageAnnotator` React component. So, it's okay.

The issue I faced was when using the hook:

```
  const anno = useAnnotator();

  useEffect(() => {
    if (anno) anno.setDrawingEnabled(editMode);
  },[anno, editMode]);
```

TypeScript was pointing that `setDrawingEnabled` doesn't exist on type

rsimon:
Yes, you'd need to type the annotator. Should be something like this:

```tsx
cons anno = useAnnotator<AnnotoriousImageAnnotator>();
```

Would have to double-check the name. (It's not called `ImageAnnotator` to avoid the clash with the React component of the same name...)

PR #525, opened 2025-03-13
Added more rect edge controls. Proper cursors. Full edge to resize.

vidya-talapady-percipient-ai:
This PR has the following changes.
*   **Edge Dragging for Rectangles:**  Rectangular annotations can now be resized by dragging directly on their edges, providing a more intuitive resizing experience. The full edge can be used for resizing.
*   **Automatic Cursor Updates:** The appropriate resize cursors (e.g., `ew-resize`, `ns-resize`, `nwse-resize`, `nesw-resize`) are automatically displayed when hovering over the edges or handles of a rectangle, giving a clear visual indication of available resizing actions.

### DIFF
--- a/src/tools/rectangle/EditableRect.js
+++ b/src/tools/rectangle/EditableRect.js
@@ -11,8 +11,11 @@ import {
   setRectMaskSize
 } from '../../selectors/RectFragment';
 
-const CORNER = 'corner';
+const HANDLE = 'handle';
 const EDGE = 'edge';
+
+const DEFAULT_EDGE_PADDING = 5;
+const DEFAULT_HANDLE_PADDING = 10;
 
 /**
  * An editable rectangle shape.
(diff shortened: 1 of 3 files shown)


src/tools/rectangle/EditableRect.js
```diff
@@ -48,128 +49,310 @@ export default class EditableRect extends EditableShape {
...
     // Mouse xy offset inside the shape, if mouse pressed
     this.mouseOffset = null;
+
+    this.svgRoot = this.svg.closest('svg');
+  }
+
+  calculateEdgePadding = () => {
+    const basePadding = this.config.edgePadding ?? DEFAULT_EDGE_PADDING;
+
+    let scaleFactor;
```
rsimon: Not 100% sure. But I think you could avoid pretty much all of this by using an SVG with a bigger stroke width (2x padding), and then using the CSS [vector-effect: non-scaling-stroke rule](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/vector-effect) to keep the size of the edge independent of the zoom factor.

vidya-talapady-percipient-ai: Could not use `vector-effect` here but the transparent padding for the handles and the edges can be controlled through css using `stroke-width`

PR #550, opened 2025-06-12
Added line shape

alexandrep-explorai:
Opening this PR following the discussion in #548 

I also wrote a tool and an editor in order to test the changes, for which I opened [this PR](https://github.com/annotorious/annotorious-plugin-tools/pull/3).

### DIFF
--- a/packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte
+++ b/packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte
@@ -47,8 +47,8 @@
 
   const onPointerMove = (canvas: HTMLCanvasElement) => (evt: PointerEvent) => {
     const {x, y} = getImageXY(new OpenSeadragon.Point(evt.offsetX, evt.offsetY));
-    
-    const hit = store.getAt(x, y, filter);
+    const scale = stage.getScale();
+    const hit = store.getAt(x, y, filter, scale);
     if (hit) {
       canvas.classList.add('hover');
 
@@ -79,7 +79,8 @@
 
     if (dist < 5) {
       const {x, y} = getImageXY(evt.position);
-      const annotation = store.getAt(x, y, filter);
+      const scale = stage.getScale();
+      const annotation = store.getAt(x, y, filter, scale);
 
       if (annotation) {
         dispatch('click', { originalEvent, annotation });
--- a/packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte
+++ b/packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte
@@ -1,10 +1,10 @@
 <script lang="ts" generics="I extends Annotation, E extends unknown">
-  import { simplifyMultiPolygon, simplifyPolygon } from '@annotorious/annotorious/src';
+  import { simplifyLine, simplifyMultiPolygon, simplifyPolygon } from '@annotorious/annotorious/src';
   import { createEventDispatcher, onMount } from 'svelte';
   import OpenSeadragon from 'openseadragon';
   import type { Annotation, DrawingStyleExpression, StoreChangeEvent, Update } from '@annotorious/core';
   import { isImageAnnotation, ShapeType } from '@annotorious/annotorious';
-  import type { Filter, ImageAnnotation, ImageAnnotatorState, MultiPolygon, Polygon } from '@annotorious/annotorious';
+  import type { Filter, ImageAnnotation, ImageAnnotatorState, Line, MultiPolygon, Polygon } from '@annotorious/annotorious';
   import type { PixiLayerClickEvent } from './PixiLayerClickEvent';
   import { createStage } from './stageRenderer';
 
@@ -40,6 +40,8 @@
     return viewer.viewport.viewportToImageCoordinates(x, y);
   }
 
+  const getIntersectionBuffer = () => 2 / stage.getScale();
+
   const onCanvasPress = (evt: OpenSeadragon.CanvasPressEvent) => {
     const { x, y } = evt.position;
     lastPress = { x, y };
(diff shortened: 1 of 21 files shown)


rsimon:
Thanks, this looks good at first glance! I'll try to take a closer look & test over the weekend. Two follow up questions:
- At least in the long run, the core should rather have a `PolyLine` implementation rather than just a `Line`. The generic polyline would then be the generic primitive for rendering polylines, simple lines (i.e. this one–from A to B), and Freehand shapes. I can take this PR for now, but it would definitely have to be replace/extend later.
- Did you test the intersection behavior with OpenSeadragon? I'm wondering whether the tolerance level needs to be made aware of the current viewport scale...

alexandrep-explorai:
- You are 100% right, having a more generic line implementation that can handle as many points as you desire would be more scalable down the line to be able to handle many kinds of line tools. I went the simpler route for this as my use case was mainly a straight line from point A to B. Since the type of the shape already handles the points as a fixed sized array (and a line should always have at the very least 2 points), it shouldn't be too much of a breaking change later if we want to extend the implementation.

- As for OpenSeaDragon, I indeed noticed I needed to be more precise when I tested the line implementation there, so you might be right that the intersects logic would need to know about the viewport scale if we want to go this route. I'll try to do more tests for that.

alexandrep-explorai:
Can confirm, at the smallest scale in OSD the line is barely selectable.

alexandrep-explorai:
@rsimon Made a new commit to make the intersection util code aware of scale so selecting in OSD works better now, also bumped the tolerance at 1.5 pixels to better reflect the actual styling of the lines.

packages/annotorious-openseadragon/src/annotation/pixi/stageRenderer.ts
```diff
@@ -426,6 +426,7 @@ export const createStage = (viewer: OpenSeadragon.Viewer, canvas: HTMLCanvasElem
   return {
     addAnnotation,
     destroy,
+    getScale: () => lastScale,
```
alexandrep-explorai, line 438: Using the last calculated scale so we don't need to constantly recalculate it in the PixiLayer, which needs it in `onPointerMove`

alexandrep-explorai:
It looked like my new commit for the OSD changes worked better at first view, but I'm having problems when testing with perfectly horizontal or vertical lines, where those lines are basically unselectable. Not quite sure why that happens.

rsimon:
No worries - let's hold off for now. This does drill a bit down into the spatial index (as you already saw) and I have some ideas how I'd like to organize things. I'll try to look into it over the weekend.

alexandrep-explorai:
Alright! Thanks a lot for looking at this!

alexandrep-explorai:
@rsimon Any updates on this PR after the weekend ? No worries if you didn't have the time to look at it, just wondering if I can help with anything.

On my side I had a quick and dirty fix I had in mind to make the feature work properly on OSD that I could implement to get a temporary version that I can use in my project for the time being.

packages/annotorious/src/model/w3c/svg/SVGSelector.ts
```diff
@@ -123,6 +151,10 @@ export const serializeSVGSelector = (shape: Shape): SVGSelector => {
   } else if (shape.type === ShapeType.MULTIPOLYGLON) {
     const geom = shape.geometry as MultiPolygonGeometry;
     value = `<svg>${serializeMultiPolygon(geom)}</svg>`;
+  } else if (shape.type === ShapeType.LINE) {
```
yuhec: We could use a switch for these if/else

yuhec (approved):

packages/annotorious/src/model/core/line/Line.ts
```diff
@@ -0,0 +1,15 @@
+import type { Bounds, Geometry, Shape } from '../Shape';
+
+export interface Line extends Shape {
+
+  geometry: LineGeometry;
+
+}
+
+export interface LineGeometry extends Geometry {
```
rsimon, line 9: As discussed, all of the core types should ultimately become `LineString` and allow for multiple segments, with the "line" simply being a special case. But let's leave it as is for now.

packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte
```diff
@@ -79,7 +79,8 @@
 
     if (dist < 5) {
       const {x, y} = getImageXY(evt.position);
-      const annotation = store.getAt(x, y, filter);
+      const scale = stage.getScale();
+      const annotation = store.getAt(x, y, filter, scale);
```
rsimon: I think `scale` seems a bit odd as a concept in the store. Might be nitpicking (and I'm fine with keeping things as they are right now!) But I guess I'd rather introduce a factor called `tolerance` or `buffer` that would simply provide a pixel distance that should be included in the intersection test. It would then be up to the consuming component (like `PixiLayer` here) to translate the scale into a reasonable `tolerance`. It's not a huge difference but IMO feels like a cleaner separation of concerns.

packages/annotorious/src/model/core/line/lineUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const LineUtil: ShapeUtil<Line> = {
+
+  area: (_: Line): number => 0,
+
+  intersects: (l: Line, x: number, y: number, scale: number = 1): boolean => {
+    const [[x1, y1], [x2, y2]] = l.geometry.points;
+
+    // Maximum distance in pixels (adjusted with scale) at which we consider the point intersects with the line
+    const tolerance = 1.5 / scale;
```
rsimon: Like above, I think exposing a `tolerance` or `buffer` arg may make more sense, semantically. This would also give the consuming component more control and allow use of different values under different circumstances. (E.g. you'll want a much higher value here for a touch device.)

rsimon (approved):

packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte
alexandrep-explorai: No no you're right, I feel like it's probably cleaner to have the scale knowledge remain in the layer where it's actually relevant, and the intersection logic can be agnostic of that concept. I adjusted my implementation in that direction.

alanliu02:
The line shape became invisible when using a large canvas, maybe setting the stroke-width will help

PR #551, opened 2025-06-13
Added `isW3CAnnotation` type guard

oleksandr-danylchenko:
## Issue
There's a relatively common need to check whether an `annotation` I'm working on uses the W3C or the internal model. Unfortunately, using just the `'@context' in annotation` isn't enough in the runtime. As during the parsing, it gets carried over to the internal model.

CleanShot 2025-06-13 at 16 56 11

## Changes Made
I added a shared `isW3CAnnotation` to conveniently differentiate between the model in the runtime.

### DIFF
--- a/packages/annotorious-core/src/model/W3CAnnotation.ts
+++ b/packages/annotorious-core/src/model/W3CAnnotation.ts
@@ -1,4 +1,4 @@
-import type { AnnotationBody } from './Annotation';
+import type { Annotation, AnnotationBody } from './Annotation';
 
 export interface W3CAnnotation {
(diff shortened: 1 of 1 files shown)


rsimon:
Thanks that looks like a generally useful addition! I'm wondering if `creator` is the best way to do it though? Perhaps `body`/`bodies` might be a useful condition, too?

oleksandr-danylchenko:
> Perhaps body/bodies might be a useful condition, too?

That makes sense! I extended the condition to look for both the `creator` and `body`

Issue #554, opened 2025-06-23
Deleting a selection of multiple points in the polygon editor allows creating invalid polygon shapes


alexandrep-explorai:
**Annotorious version**
`@annotorious/annotorious`, version 3.4.7 (latest)

**Issue behavior**

The polygon editor has a guard in place when you try to delete a point and the polygon only has 3 remaining points. 

The problem with this guard is that it does not take into account that the editor allows selecting multiple points and deleting them in a batch. Doing so when the polygon has more than 3 points allows to save a polygon with 2, 1, or even no points (this last case even causing a runtime exception).

Video demonstrating the issue:
https://github.com/user-attachments/assets/287b0ad7-e724-4fd4-b74b-48b7d9a52da7

Two points polygon:
<img width="528" alt="Image" src="https://github.com/user-attachments/assets/b6742d7d-effa-4e10-9d3e-49e4ffde30c1" />

One point polygon:
<img width="528" alt="Image" src="https://github.com/user-attachments/assets/c2b2fb3b-0260-4e80-9e22-f4b9947ddb15" />

Zero point polygon:
<img width="557" alt="Image" src="https://github.com/user-attachments/assets/d33656e7-a146-4e51-9de7-170437b5e5a9" />

rsimon:
Thanks for the PR–merged! New release v3.5.0 of Annotorious is out on npm 🎉

PR #555, opened 2025-06-24
Addresses #554

Flecks1:
Simply made it so we make sure that the amount of points in the polygon (or the ring in the case of a multipolygon shape) remains at a minimum of three when we subtract the amount of point we are deleting.


rsimon:
Wonderful–thanks!

PR #589, opened 2025-11-10
Make @annotorious/openseadragon a peerDependency of @annotorious/react

### LINKED ISSUE #588: @annotorious/openseadragon should be an optional peerDependency of @annotorious/react
## Problem
The `@annotorious/openseadragon` package is currenctly a hard dependency of the `@annotorious/react` package, even if the consumer has no desire to use the specific components that leverage `@annotorious/openseadragon`. Being forced to install `@annotorious/openseadragon` has a negative impact on bundle size, especially because that also requires the consumer to install the `openseadragon` package.

## Proposed Solution
From my understanding, the base `@annotorious/annotorious` package does not require `openseadragon` by default. Instead it only requires `openseadragon` when leveraging `openseadragon` functionality.

Ideally the `@annotorious/react` package should work the exact same way.

What I am proposing here from the consumer perspective is that if the consumer just wants the base `@annotorious/react` components then that should be possible with the following import
```tsx
// Does not require openseadragon or @annotorious/openseadragon.
import { Annotorious } from '@annotorious/react'; 
```

And if the consumer wants one of the openseadragon components then they can import it like this
```tsx
// Requires openseadragon and @annotorious/openseadragon 
import { OpenSeadragonAnnotator } from '@annotorious/react/openseadragon';
```

blchelle:
## Problem
The `@annotorious/openseadragon` package is currenctly a hard dependency of the `@annotorious/react` package, even if the consumer has no desire to use the specific components that leverage `@annotorious/openseadragon`. Being forced to install `@annotorious/openseadragon` has a negative impact on bundle size, especially because that also requires the consumer to install the `openseadragon` package.

## Changes
- Move all of the `openseadragon` dependent code into a separate export `@annotorious/react/openseadragon`
- Update the vite config to accomodate this this change.

Note: I believe this will likely require a major version upgrade since we're changing the import path for all of the OpenSeaDragon related components.

Closes #588


packages/annotorious-react/package.json
```diff
@@ -35,19 +40,18 @@
     "vite-tsconfig-paths": "^5.1.4"
   },
   "peerDependencies": {
-    "openseadragon": ">= ^4.0.0 || >= ^5.0.0",
+    "@annotorious/openseadragon": "3.7.14-beta",
```
blchelle, line 43: I am not certain what versions of `@annotorious/openseadragon` we want to enforce for the peerDependency.

blchelle:
Going to close this since it seems like it's not going to get any action.

AndrewADev:
@rsimon since this was re-opened: what would be needed to get these or similar changes merge-ready? (just updates to latest base? or are there also specific changes you had in mind?)

Thanks in advance!

PR #597, opened 2026-02-12
fix: incorrect stroke widths with multiple OSD viewers

### LINKED ISSUE #596: PixiJS stroke widths incorrect with multiple OpenSeadragon viewers due to shared module-level `lastScale`
## Bug Description

When using multiple `OpenSeadragonAnnotator` instances on the same page (e.g. side-by-side image comparison), annotation stroke widths render incorrectly on one viewer after interacting with the other. Specifically, hovering over annotations on viewer B after zooming viewer A causes the annotation outlines to appear too thick or too thin.

## Root Cause

In `packages/annotorious-openseadragon/src/annotation/pixi/stageRenderer.ts`, the variables `lastScale` and `fastRedraw` are declared at **module scope** (lines 31-34):

```typescript
let fastRedraw = false;
let lastScale: number;
```

These are shared across all `createStage` instances. Each viewer's `redrawStage` closure overwrites `lastScale` with its own viewport scale.

The `drawShape` higher-order function (line 97) uses `lastScale` to compute stroke width:

```typescript
const lineWidth = strokeStyle.lineWidth === 1 ? 1 : strokeStyle.lineWidth / lastScale;
```

When `setHovered` triggers `redrawAnnotation` → `addAnnotation` → `drawShape`, it reads `lastScale` which may have been set by a **different** viewer's `redrawStage`. This produces incorrect stroke widths.

## Steps to Reproduce

1. Render two `OpenSeadragonAnnotator` instances side-by-side, each with annotations using `strokeWidth > 1`
2. Zoom in/out on viewer A
3. Hover over an annotation on viewer B
4. Observe: the annotation outline appears at the wrong thickness (too thick or too thin depending on relative zoom levels)

## Proposed Fix

Sync `lastScale` to the current viewer at the start of `addAnnotation`, before `drawShape` reads it:

```typescript
const addAnnotation = (annotation: ImageAnnotation, state?: AnnotationState) => {
    lastScale = getCurrentScale(viewer);  // ensure correct scale for this viewer
    fastRedraw = false;
    // ... rest unchanged
```

This is a minimal fix. A more thorough solution would be to move `lastScale` and `fastRedraw` into `createStage`'s closure and pass them to `drawShape`, eliminating the shared mutable state entirely.

## Environment

- `@annotorious/openseadragon`: 3.7.19
- `@annotorious/react`: 3.7.19

casimp:
## Summary

Fixes #596 — annotation stroke widths render incorrectly when multiple `OpenSeadragonAnnotator` instances share a page.

## Problem

`lastScale` and `fastRedraw` are declared at module scope in `stageRenderer.ts` (lines 31–34) and shared across every `createStage` instance. When viewer A's `redrawStage` writes its viewport scale to `lastScale`, a subsequent `addAnnotation` call on viewer B (e.g. triggered by hover) reads A's scale value. `drawShape` then computes `strokeWidth / lastScale` with the wrong denominator, producing visibly incorrect stroke widths.

## Fix

Sync `lastScale` to the current viewer at the top of `addAnnotation`, before any shapes are drawn:

```typescript
const addAnnotation = (annotation: ImageAnnotation, state?: AnnotationState) => {
    lastScale = getCurrentScale(viewer);
    fastRedraw = false;
    // ...
```

This is a minimal, targeted fix. A more thorough follow-up could move `lastScale` and `fastRedraw` into `createStage`'s closure to eliminate the shared mutable state entirely.

## Test plan

- [ ] Render two `OpenSeadragonAnnotator` instances side-by-side, both with annotations using `strokeWidth > 1`
- [ ] Zoom in/out on viewer A
- [ ] Hover over annotations on viewer B — stroke widths should remain correct
- [ ] Verify single-viewer usage is unaffected

### DIFF
--- a/packages/annotorious-openseadragon/src/annotation/pixi/stageRenderer.ts
+++ b/packages/annotorious-openseadragon/src/annotation/pixi/stageRenderer.ts
@@ -297,8 +297,15 @@ export const createStage = (viewer: OpenSeadragon.Viewer, canvas: HTMLCanvasElem
   lastScale = getCurrentScale(viewer);
 
   const addAnnotation = (annotation: ImageAnnotation, state?: AnnotationState) => {
+    // Ensure lastScale reflects this viewer's current viewport scale.
+    // lastScale is module-scoped and shared across all createStage instances,
+    // so it may hold a stale value from a different viewer. drawShape reads
+    // lastScale to compute counter-scaled stroke widths, so we must sync it
+    // here before any shapes are drawn.
+    lastScale = getCurrentScale(viewer);
+
     // In case this annotation adds stroke > 1
-    fastRedraw = false; 
+    fastRedraw = false;
 
     // Robustness: make sure to delete any annotation with the same ID, because 
     // otherwise we'd replace the annotation in `annotationShapes` and leave 


rsimon:
Good catch - and yes, that looks like a good patch for now.

FWIW: the whole rendering logic is due for a rewrite IMO. It's grown too much, and frankly carries too much legacy baggage: Such as forced compatibility with the SVG-only implementation from `@annotorious/annotorious`; various hacks as I introduced while stumbling through multiple major version upgrades of PixiJS; and an overall "Pixi-unfriendly" architecture that's not using a standard render loop (and this mostly due to my own initial lack of knowledge of how Pixi is supposed to work.)

Issue #603, opened 2026-05-21
Potential ReDoS and Strict Parser Bypass in W3C FragmentSelector Parsing


WiiiiillYeng:
Hi, I would like to report a potential security issue in the W3C `FragmentSelector` parsing logic. I have tried to keep this report limited to the observable behavior and a minimal reproduction.

## Summary

The `xywh` media fragment parser appears to have two related issues:

1. A crafted `FragmentSelector.value` can cause superlinear CPU usage when parsed through the public W3C annotation parser path.
2. `strict: true` accepts malformed `xywh` coordinate values such as `10abc`, because they are parsed with `parseFloat` and silently converted to numeric prefixes.

The ReDoS case is the more security-relevant issue. The strict parsing issue is included because it appears to be caused by the same parser accepting overly broad coordinate fields.

## Affected API

The behavior is reachable through public APIs:

```js
W3CImageFormat(source, { strict: true, invertY: false }).parse(annotation)
parseFragmentSelector(fragment)
```

I tested this with the installed package:

```text
@annotorious/annotorious 3.8.2
```

## Trigger Conditions

An attacker would need to control W3C annotation data that is later parsed by an application using Annotorious. For example:

- a user-uploaded W3C annotation JSON file,
- annotations loaded from a remote annotation service,
- collaborative annotation data created by another user,
- server-side or client-side import of third-party annotation data.

The relevant attacker-controlled field is a `FragmentSelector` value, for example:

```json
{
  "type": "Annotation",
  "target": {
    "type": "SpecificResource",
    "selector": {
      "type": "FragmentSelector",
      "value": "..."
    }
  }
}
```

## Potential Impact

If untrusted annotation data is parsed, the ReDoS payload can block the JavaScript thread during parsing.

Possible consequences:

- browser UI freeze or noticeable page hangs when parsing annotations on the main thread,
- Node.js event loop blocking if annotations are parsed server-side,
- denial of service against annotation import or viewing workflows.

The strict parsing behavior may also allow malformed selector values to be accepted as valid rectangles. This may cause invalid or non-standard annotation data to be imported, stored, rendered, or passed to downstream systems that expect strict W3C media fragment parsing.

## Proof of Concept

The following script imports the real package and exercises the public parser path.

```js
import { performance } from 'node:perf_hooks';
import { createRequire } from 'node:module';
import { W3CImageFormat, parseFragmentSelector } from '@annotorious/annotorious';

const require = createRequire(import.meta.url);

[...]

PR #604, opened 2026-05-24
Anchored regex + strict numeric parsing for W3C FragmentSelector

truffle-dev:
## In this PR

Addresses #603.

The W3C `FragmentSelector` parser used a non-anchored regex with broad
lazy groups and a nested quantifier (`(.+?),(.+?),(.+?),(.+)*`). On
crafted inputs like `'xywh='.repeat(n) + '\nxywh=1,2,3,4'` the engine
backtracked superlinearly: at `n=16000` the issue reporter measured
parse times around 485 ms. The coordinate fields were also parsed with
`parseFloat`, which silently truncated `'10abc'` to `10`, so values
like `xywh=pixel:10abc,20def,30ghi,40jkl` were accepted as a valid
rectangle even in `strict: true` mode.

* `FragmentSelector.ts` now uses a single anchored, numeric-only
  expression (`^xywh=(?:(pixel|percent):)?N,N,N,N$`) where `N` allows
  integers, decimals, scientific notation, and a leading minus. URI-
  style inputs (`https://.../image.jpg#xywh=...`) are normalized via a
  small `stripHash` helper before matching, so the regex itself never
  scans across arbitrary URI prefixes.
* Coordinate fields are now converted with `Number`, which rejects
  non-numeric suffixes (`10abc` → `NaN`); since the regex no longer
  matches such tokens, the parser throws `Not a MediaFragment` instead
  of returning a half-truncated rectangle.

The existing five tests still pass unchanged. Four new tests cover:

* malformed coordinate rejection (`10abc,...`),
* bounded parse time on the reporter's adversarial payload at n=16000,
* decimal coordinates,
* URI fragments with `#xywh=`.

On my machine the reporter's largest payload (n=64000, ~320 KB) now
parses in well under a millisecond.

### DIFF
--- a/packages/annotorious/src/model/w3c/fragment/FragmentSelector.ts
+++ b/packages/annotorious/src/model/w3c/fragment/FragmentSelector.ts
@@ -11,6 +11,23 @@ export interface FragmentSelector {
 
 }
 
+const NUMBER = '-?\\d+(?:\\.\\d+)?(?:[eE][+-]?\\d+)?';
+
+// Anchored, numeric-only pattern. Avoids the broad lazy + nested quantifier
+// shapes that allowed pathological backtracking on adversarial input.
+const XYWH_RE = new RegExp(
+  `^xywh=(?:(pixel|percent):)?(${NUMBER}),(${NUMBER}),(${NUMBER}),(${NUMBER})$`,
+  'i'
+);
+
+// Strip everything up to and including the first '#xywh=' for URI-style inputs
+// like "https://example.org/image.jpg#xywh=...". Returns the bare fragment when
+// no hash is present.
+const stripHash = (s: string): string => {
+  const idx = s.indexOf('#xywh=');
+  return idx < 0 ? s : s.slice(idx + 1);
+}
+
 export const isFragmentSelector = (
   selector: any
 ): boolean => {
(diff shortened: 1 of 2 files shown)


WiiiiillYeng:
I believe replacing the broad `(.+?)` groups with the explicit numeric pattern `-?\\d+(?:\\.\\d+)?(?:[eE][+-]?\\d+)?` is sufficient to address the issue. Adding `^` to anchor the regex is not necessary here, because it would require additional preprocessing to avoid missing valid `xywh=` fragments that do not start at the beginning of the input.

truffle-dev:
Right, the numeric pattern alone closes the ReDoS — without the broad lazy groups the engine has no ambiguity to backtrack through. The anchoring isn't load-bearing for the security fix.

I anchored anyway for a separate reason: W3C defines the FragmentSelector value as the full `xywh=...` string. Unanchored, the parser accepts `xywh=1,2,3,4` embedded in arbitrary surrounding text, which is broader than `strict: true` should allow. The `stripHash` helper handles the one realistic non-anchored case (URI fragments like `image.jpg#xywh=...`) explicitly.

Happy to drop anchoring + `stripHash` and lean on the numeric pattern alone if the maintainer prefers the narrower change. Either shape closes the issue you reported.

truffle-dev:
Superseded by 61148bf (v3.8.3) — the narrower numeric regex (matching @WiiiiillYeng's suggestion) plus the 512-char length cap close it cleanly. Closing.

PR #607, opened 2026-05-25
Added `syncAnnotations` method with diff-based reconciliation that preserves active selection

oleksandr-danylchenko:
## Issue
Fixes https://github.com/annotorious/annotorious/issues/605

## Changes Made
`setAnnotations` previously performed a destructive wipe-and-reinsert when called with `replace=true` (the default), reporting every existing annotation as deleted. Even if the latter reappeared in the new payload, causing active selections to be dismissed mid-interaction. 

This PR introduces diff-based reconciliation: annotations present in both the old and new state are updated in-place, new ones are added, _and only annotations genuinely absent from the new payload are deleted_. That was the key piece missing from the existing store methods infrastructure.

As a side effect, a repeated index-manipulation pattern across the store was consolidated into a shared helper. 

The result is that periodic data refreshes no longer interrupt the user — open popups and editors stay mounted as long as the selected annotation is still included in the incoming data.

### DIFF
--- a/packages/annotorious-core/src/model/Annotator.ts
+++ b/packages/annotorious-core/src/model/Annotator.ts
@@ -46,13 +46,13 @@ export interface Annotator<I extends Annotation = Annotation, E extends unknown 
 
   getUser(): User;
 
-  loadAnnotations(url: string, replace?: boolean): Promise<E[]>;
+  loadAnnotations(url: string): Promise<E[]>;
 
   redo(): void;
 
   removeAnnotation(arg: Partial<E> | string): E | undefined;
 
-  setAnnotations(annotations: Partial<E>[], replace?: boolean): void;
+  setAnnotations(annotations: Partial<E>[]): void;
 
   setFilter(filter: Filter<I> | undefined): void;
 
@@ -136,11 +136,11 @@ export const createBaseAnnotator = <I extends Annotation, E extends unknown>(
       : selected as unknown as E[];
   }
 
-  const loadAnnotations = (url: string, replace = true) =>
+  const loadAnnotations = (url: string) =>
     fetch(url)
       .then((response) => response.json())
       .then((annotations) => {
-        setAnnotations(annotations, replace);
+        setAnnotations(annotations);
         return annotations;
       });
 
@@ -161,7 +161,7 @@ export const createBaseAnnotator = <I extends Annotation, E extends unknown>(
     }
   }
 
-  const setAnnotations = (annotations: E[], replace = true) => {
+  const setAnnotations = (annotations: E[]) => {
     if (adapter) {
       const parseFn = adapter.parseAll || parseAll(adapter);
       const { parsed, failed } = parseFn(annotations);
@@ -169,18 +169,9 @@ export const createBaseAnnotator = <I extends Annotation, E extends unknown>(
       if (failed.length > 0)
         console.warn(`Discarded ${failed.length} invalid annotations`, failed);
 
-      if (replace) {
-        store.syncAnnotations(parsed, Origin.REMOTE);
-      } else {
-        store.bulkAddAnnotations(parsed, false, Origin.REMOTE);
-      }
+      store.syncAnnotations(parsed, Origin.REMOTE);
     } else {
-      const mapped = annotations.map(reviveDates<I>);
-      if (replace) {
-        store.syncAnnotations(mapped, Origin.REMOTE);
-      } else {
-        store.bulkAddAnnotations(mapped, false, Origin.REMOTE);
-      }
+      store.syncAnnotations(annotations.map(reviveDates<I>), Origin.REMOTE);
     }
   }
 
--- a/packages/annotorious-core/src/model/Annotator.ts
+++ b/packages/annotorious-core/src/model/Annotator.ts
@@ -169,9 +169,18 @@ export const createBaseAnnotator = <I extends Annotation, E extends unknown>(
       if (failed.length > 0)
         console.warn(`Discarded ${failed.length} invalid annotations`, failed);
 
-      store.bulkAddAnnotations(parsed, replace, Origin.REMOTE);
+      if (replace) {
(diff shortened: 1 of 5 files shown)


rsimon:
Thanks - this looks like a great & sensible addition!

I guess my only "concern" is this: with the introduction of `syncAnnotation`, do we even need the old `bulkAddAnnotations` method anymore at all? Feels like `syncAnnotation` does what `bulkAddAnnotations` should have done from the start, perhaps?

oleksandr-danylchenko:
Hello, @rsimon 👋🏻 
Sorry for the long response. I missed your comment in my notifications.

---

That's a great question! After I gave it another look, I'm not sure why someone would prefer the "non-replace" `bulkAddAnnotations` method. It will essentially "append" any passed annotations. However, if a user passes an annotation with the same `id`, it may cause havoc. Instead, the `syncAnnotations` is a safer option that will still only append if all the annotations are brand-new.

Therefore, we may be able to get rid of that additional `replace` argument and the `bulkAddAnnotations` in general 🚀

oleksandr-danylchenko:
Now you should be able to get another look at the PR 🙏🏻

rsimon:
Wait - I might be misunderstanding this. But the `replace` flag is important. If set to `true`, the function should indeed clear all existing annotations (unless they are **also** in the new list). I.e. it works more like a "true" set function rather than a merge functions if that makes sense?

oleksandr-danylchenko:
> If set to true, the function should indeed clear all existing annotations (unless they are also in the new list). 

Yeah, that's the idea behind the `syncAnnotations` method. Any annotations that are missing from the new passed list will be removed. So it's not a "merge" per se. 
Instead, the `bulkUpsertAnnotations` performs the "merge", by only adding or updating annotations.

I think I was too "heavy-handed" by removing the `replace`. The alt path (`replace=false`) should have called the `bulkUpsertAnnotations`.

rsimon (approved):
Got it - yes, looks lovely now! Thanks so much for this!

oleksandr-danylchenko:
Thanks for the merge, @rsimon!
I also made an update for the `annotorious-website` to include documentation about the updated behavior of the `setAnnotations` with the `syncAnnotations` under the hood - https://github.com/annotorious/annotorious-website/pull/4

rsimon:
Ah - thanks, I don't know why I missed the update to the website docs! Merged & online now!

oleksandr-danylchenko:
I realized that a bunch of code depending on the `bulkAddAnnotations` should be updated in the `text-annotator-js`. Also, the previous `AnnotatorPlugin` definition strategy should be updated accordingly[^1].

Currently, in my app-specific plugin, I override the `addAnnotation`, `bulkAddAnnotations`, `updateAnnotation`, and `updateTarget` with a custom crosswalk layer. Instead, the `bulkAddAnnotations` override should be replaced with the `syncAnnotations` and `bulkUpsertAnnotations` overrides.

[^1]: You mentioned how the TEI plugin was monkey-patching the store methods: https://github.com/oleksandr-danylchenko/text-annotator-js/commit/d24edf73daf034bc7b080d481c7f38550c99fc4d#commitcomment-188407106

rsimon:
I kept the `bulkAddAnnotations` method with a `@deprecated` annotation, so everything should continue working normally (fingers crossed). But, yes, code should be updated to use either `bulkUpsert` or `sync` directly.

oleksandr-danylchenko:
> I kept the `bulkAddAnnotations` method with a `@deprecated` annotation, so everything should continue working normally (fingers crossed). But, yes, code should be updated to use either `bulkUpsert` or `sync` directly.

In my project, I think it would be worth taking a closer look at the https://github.com/recogito/text-annotator-js/commit/d5b28933acae6ecb65fe1038eefd361acb3cf4bc 🤔 
There, the whole plugin layer was replaced with a single `reviveSelectorFn`. That seems like a more robust approach compared to the monked-patching. Especially since the `Store` API is prone to change and I've already missed updating the overrides accordingly.

Then, there's another dilemma of what should be overridden and what should be left as-is. Which can be painful to decide for the methods that might invoke each other. E.g. "should I apply the crosswalk in the `syncAnnotations` override? It _might_ call the `addAnnotation` under the hood. But there's no way of knowing that w/o looking at the sources 🤷🏻‍♂️ "

rsimon:
Exactly! I feel your pain, and it's exactly like the TEI plugin worked before. Now, everything is "collapsed" into that single function. Felt perfectly elegant to me - until I realized the "revive" is actually doing two seperate things underneath the hood.

I'm still inclined to keep it that way, simply because it works well in practice. (Inside the store functions that use reviving *do* get called from both "outside" and "inside", i.e. via the API as well as driven by user select/create actions. So the function is already inherently facing two different inputs. Feels like a code smell, but I have no idea how to untangle that without a deeper rethink of the architecture.)

oleksandr-danylchenko:
Ohh... I noticed a new limitation in the `selectorReviveFn`. It expects that only _a single_ `RevivedTextSelectorLike` will be returned ([source](https://github.com/recogito/text-annotator-js/blob/60c092d99a799eaa3958667103622219e25e88aa/packages/text-annotator/src/text-annotator-options.ts#L13-L17)). However, in my app, _the single initial selector can be split into multiple ones_, depending on what content is covered. I can extend that 👌🏻

rsimon:
> Ohh... I noticed a new limitation in the `selectorReviveFn`. It expects that only _a single_ `RevivedTextSelectorLike` will be returned ([source](https://github.com/recogito/text-annotator-js/blob/60c092d99a799eaa3958667103622219e25e88aa/packages/text-annotator/src/text-annotator-options.ts#L13-L17)). However, in my app, _the single initial selector can be split into multiple ones_, depending on what content is covered.

Oh - got it. Yes, but that seems like a general case that extensions may want to do. Happy to take that case into consideration for an interface change!

oleksandr-danylchenko:
The multi-selector splitting concern was addressed in the https://github.com/recogito/text-annotator-js/pull/277 🤖

PR #609, opened 2026-05-28
Attach presence key to globalThis

blms:
Implementing the suggestion here: https://github.com/recogito/recogito-studio-sdk/pull/7#issuecomment-4562767233

Let me know if this is what you had in mind, or feel free of course to scrap/make adjustments!


rsimon:
Thanks - looks good!

PR #610, opened 2026-06-11
Make OpenSeadragon polygon simplification tolerance configurable

### LINKED ISSUE #595: Bug: Gratuitous polygon simplification
Hello! I have some complex polygons. When I am editing them, the detailed shape is preserved, but when I deselect-  all details are lost. Is there a way to disable this? If not, can you point me to the code where this is happening, so I could make the changes and maybe PR it?

(simplified)
Image

(complex)
Image

(Is it due  to this?
https://github.com/annotorious/annotorious/blob/32cdd23a7a75a5ff207d918bf7390297f082fe9b/packages/annotorious-openseadragon/src/annotation/pixi/PixiLayer.svelte#L190 )

VrtxOmega:
Closes #595.

## Summary

- add a `polygonSimplificationTolerance` OpenSeadragon option for the Pixi display layer
- keep the existing default tolerance of `1` so current rendering behavior stays unchanged by default
- allow callers to set the tolerance to `0` when they need full polygon/multipolygon geometry preserved
- add regression coverage for polygon and multipolygon simplification tolerance

## Validation

- `npm test --workspace @annotorious/annotorious -- simplificationTolerance.test.ts`
- `npm test --workspace @annotorious/annotorious`
- `npm run build --workspace @annotorious/openseadragon`
- `npm test`
- `npm run build`

### DIFF
--- a/packages/annotorious-openseadragon/src/Annotorious.ts
+++ b/packages/annotorious-openseadragon/src/Annotorious.ts
@@ -115,7 +115,8 @@ export const createOSDAnnotator = <I extends Annotation = ImageAnnotation, E ext
       state, 
       viewer, 
       style: opts.style,
-      filter: undefined
+      filter: undefined,
+      polygonSimplificationTolerance: opts.polygonSimplificationTolerance
     }
   });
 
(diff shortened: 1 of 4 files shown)


rsimon:
Looks good - many thanks!

VrtxOmega:
Thanks for the quick review and merge. Glad this helps OpenSeadragon users preserve full polygon geometry when needed while keeping the default behavior unchanged.

Small note: this went through my local PCF contribution gate before submission, which helped keep the change narrow and regression-tested.

PR #615, opened 2026-07-24
Fix coordinate mapping css transformed ancestors

sultaniman:
## What changed
- Map pointer coordinates via `viewBox`/bounding-box ratios instead of `getScreenCTM()`
(`Transform.ts`, `SVGAnnotationLayerPointerEvent.ts`, `Editor.svelte`), and via the
viewer's bounding rect instead of OSD's raw offsets (shared `getViewerPoint` util, used
by the Pixi and SVG drawing layers).
- Add an OpenSeadragon coordinate-mapping playground (`npm run playground:osd`) and a unit
test for `createSVGTransform` - do we need the playgroun?

## OSD playground app
Manual: `npm run playground:osd`, toggle the CSS transform, then select, drag, resize,
draw, and zoom to 2-8x.

Resolves: https://github.com/annotorious/annotorious/issues/613

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "npm run build --workspaces",
+    "playground:osd": "npm run start --workspace @annotorious/react -- --open /test/openseadragon-coordinate-mapping/index.html",
     "test": "npm run test --workspaces"
   },
   "workspaces": [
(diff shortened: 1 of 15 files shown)


rsimon:
Thanks for this. Just a heads-up that it may take a bit until I can review and test – but I really appreciate it!

